PR #29, opened 2024-01-09
🐛 FIX: Theme flicker

ahmadbilaldev:
This PR attempts to fix the theme flicker and the related hydration errors that arise due to Next.js 13 behaviour with themes.

@MaedahBatool Please take a look.


MaedahBatool:
LGTM 👍

PR #42, opened 2024-01-18
search: remove links to .com for structural search

stefanhengl:
We remove the links because structural search will be disabled on .com and the links will not work anymore.


MaedahBatool (approved):
LGTM 👍

PR #48, opened 2024-01-21
📦 NEW: Add versioning

ahmadbilaldev:
This PR adds a versioning support to docs. 

It will enable multiple versions locally added in separate folders inside `docs/versioned`. The latest docs will stay in `docs` as before. 

The link to latest docs will be `sourcegraph.com/docs/` while for versioned docs it will be `sourcegraph.com/docs/v/{versionName}/`, for instance  `sourcegraph.com/docs/v/5.2.1/`.

### Working Features:

- Updated route structure, combined 5 routes into a single catch all route. Previously Server routes are now SSG, speeding up page load times significantly. 
- Version selector dropdown to the top nav and a legacy versions page. 
- Docs/versioned folder for adding versioned docs. For example: Add a `docs/versioned/5.2.1` folder to add docs for 5.2.1
- Version list. Add active (non-legacy) versions list to `data/versions.ts`. Example below: 
	```
	export interface VersionI {
	    name: string;
	    label?: string; // Optional - Shown in the version dropdown
	    url: string;
	}
	
	// Ensure the first entry is the latest active version
	export const versions: VersionI[] = [
	    // Latest version
	    {
	        name: 'v5.2.3',
	        label: 'latest',
	        url: '/'
	    },
	    // Example version
	    {
	        name: 'v5.2.1',
	        url: '/v/5.2.1'
	    },
	];
	
	export const latestVersion = versions[0];
	```

- Dynamically prepend versions to links inside docs.
- Dynamically prepend versions to links inside MDX components. 
- Dynamic versioning support in Side Nav, Breadcrumbs.

### Caveats

Versioning is ready for production. However here are some minor caveats that can probably be fixed later:

- Searching on versioned docs will show results from latest docs. 
- Previous and next links at bottom of the page are disabled for versioned docs.

@MaedahBatool Please review the PR on your end.


MaedahBatool:
LGTM 👍

PR #81, opened 2024-02-07
Add pricing docs for Pro plan

MaedahBatool:
The PR adds docs and FAQs for Cody's Pro plan.

### DIFF
--- a/docs/cody/usage-and-pricing.mdx
+++ b/docs/cody/usage-and-pricing.mdx
@@ -68,15 +68,21 @@ We charge payments at the beginning of each billing cycle, and they get automati
 
 ### What is a billing cycle?
 
-The billing cycle refers to the start date and depends on your account creation date. You can view your current and previous billing cycles from the **My subscription** tab.
+The billing cycle refers to the start date you start your Cody Pro plan. Your Cody Pro plan will automatically renew at each billing date. You can view your current and previous billing cycles from the **My subscription** tab.
 
 ### What payment methods are available?
 
-We only support payments through a credit card.
+We currently only support payments through a credit card.
 
 ### What happens when I cannot pay or don't want to renew the Pro plan?
 
-If you are on a Pro plan trial, you must add a credit card to your account to continue on a Pro plan. If you do not want to renew Cody Pro, you will automatically downgrade to Cody Free.
+If you do not want to renew Cody Pro, you can cancel your Cody Pro subscription at any time by going to the [accounts page](https://accounts.sourcegraph.com/cody/subscriptions). If you cancel the Pro plan, your Cody Pro will continue until the end of the billing cycle. Once the billing cycle ends, your plan will not renew, and your card will not be charged. This automatically downgrades your plan from Cody Pro to Cody Free.
+
+If you change your mind after canceling your plan and you have a credit card on file, you can click **Resume subscription** from the [accounts page](https://accounts.sourcegraph.com/cody/subscriptions) to re-upgrade to Cody Pro.
+
+### Are there any refunds for the Pro subscription?
+
+We don't offer refunds, but if you have any queries regarding the Cody Pro plan, please write to support@sourcegraph.com, and we'll help resolve the issue.
 
 ## Enterprise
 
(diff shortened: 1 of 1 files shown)


docs/cody/usage-and-pricing.mdx
```diff
@@ -47,6 +56,28 @@ To revert back to Cody Free from Pro:
...
+
+You can get this info from our [pricing page](https://sourcegraph.com/pricing?product=cody).
+
+### When are payments taken?
+
+We charge payments at the beginning of each billing cycle, and they get automatically renewed once you've added your credit card details.
+
+### What is a billing cycle?
+
+The billing cycle refers to the start date and depends on your account creation date. You can view your current and previous billing cycles from the **My subscription** tab.
```
chenkc805, line 71: The billing cycle doesn't depends on the account creation date. It depends only on when the user starts their Cody Pro plan.

docs/cody/usage-and-pricing.mdx
```diff
@@ -47,6 +56,28 @@ To revert back to Cody Free from Pro:
...
+
+### What payment methods are available?
+
+We only support payments through a credit card.
```
chenkc805, line 75: Maybe change to "We currently only..." because we hope to add more in the distant future

docs/cody/usage-and-pricing.mdx
```diff
@@ -47,6 +56,28 @@ To revert back to Cody Free from Pro:
...
+
+### What happens when I cannot pay or don't want to renew the Pro plan?
+
+If you are on a Pro plan trial, you must add a credit card to your account to continue on a Pro plan. If you do not want to renew Cody Pro, you will automatically downgrade to Cody Free.
```
chenkc805, line 79: We should remove "If you are on a Pro plan trial, you must add a credit card to your account to continue on a Pro plan" on 2/21 because trials will no longer a thing anymore. 

I think we should have some information about cancelling their plan. So something like:

- If you do not want to renew Cody Pro, you can cancel your Cody Pro subscription at any time by going to accounts.sourcegraph.com/cody/subscriptions. If you cancel your Cody Pro plan, your Cody Pro plan will continue until the end of the billing cycle, at which point your plan will not renew, your card will not be charged, and your plan will change from Cody Pro to Cody Free. 
- If you change your mind after cancelling your plan and you have a credit card on file, you can simply click "Resume subscription" on accounts.sourcegraph.com/cody/subscriptions to re-upgrade to Cody Pro.

docs/cody/usage-and-pricing.mdx
```diff
@@ -47,6 +56,28 @@ To revert back to Cody Free from Pro:
...
+
```
chenkc805, line 80: We should have a section about Refunds. Something like: 

- We don't offer refunds, but if you have a problem with your Cody Pro plan, please write to [support@sourcegraph.com](mailto:support@sourcegraph.com) and we can work with you to resolve your problem.

docs/cody/usage-and-pricing.mdx
chenkc805: I think we should also reference something like "Your Cody Pro plan will automatically renew at each billing date."

docs/cody/usage-and-pricing.mdx
chenkc805: CC: @enriquegh

chenkc805 (changes requested):

chenkc805:
Thanks Maedah. Some other feedback reading through the rest of the docs:

- For this line: https://github.com/sourcegraph/docs/blob/b566dcf83125eca5519f6ad9ca5de4fde260158d/docs/cody/usage-and-pricing.mdx#L53
  - Change to "This automatically downgrades you to Cody Free once after your billing cycle finishes"

- For this line: https://github.com/sourcegraph/docs/blob/b566dcf83125eca5519f6ad9ca5de4fde260158d/docs/cody/usage-and-pricing.mdx#L103
  - Let's remove the line around embeddings. Embeddings is much less important and I don't want users to get hung up on these numbers. 
  - CC: @iskyOS what do you think about this proposal for this line?

docs/cody/usage-and-pricing.mdx
enriquegh: Agreed. The blurb Kevin added looks good

chenkc805 (changes requested):
see my above comment

MaedahBatool:
@chenkc805 done with the remaining feedback. Are we ready to merge this?

PR #115, opened 2024-02-28
Document error related to Docker not being found

varungandhi-src:


docs/dev/background-information/bazel/faq.mdx
```diff
@@ -68,6 +68,18 @@ You can also use the same configuration flag to run the container tests on MacOS
...
 
+### I am not able to run image tests locally as Docker is not detected
+
+If you get an error like:
+
+```
+time="2024-02-28T06:31:07Z" level=fatal msg="error loading oci layout into daemon: error loading image: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?, %!s(MISSING)"
+```
+
+It might be because Docker Desktop might not be configured to use the default Docker socket during installation.
```
MaedahBatool, line 79:
```suggestion
It might be because the Docker Desktop might need to be configured to use the default Docker socket during installation.
```

docs/dev/background-information/bazel/faq.mdx
```diff
@@ -68,6 +68,18 @@ You can also use the same configuration flag to run the container tests on MacOS
...
+
+Make sure that the checkbox under Docker Desktop -> Settings -> Advanced -> "Allow the default Docker socket to be used (requires password)" is enabled. If it wasn't, then enable it and restart Docker.
```
MaedahBatool, line 81:
```suggestion
Make sure that the checkbox under **Docker Desktop > Settings > Advanced > Allow the default Docker socket to be used (requires password)** is enabled and then restart Docker.
```

MaedahBatool (changes requested):
Adding small fixes.

docs/dev/background-information/bazel/faq.mdx
varungandhi-src: > the Docker Desktop

I don't think this is grammatically correct. For example: https://english.stackexchange.com/questions/188410/use-of-the-in-front-of-an-application-name

> Usage of "the" is suggested only when you use a word like app with mSecure, ex- "Applying the concepts of HCI in development of the mSecure app". However, go without the "the".

> Don't use "the" with your application name. In this case the application name functions in the same way as the name of a person (or dog, or other animal I suppose) and it would be unusual (and highly stylistic) to place a "the" in front of the name.

MaedahBatool (approved):

PR #117, opened 2024-02-28
small fix for some json files - remove dangling commas

peterguy:
A couple of json files that are used for reference had dangling commas in them.


MaedahBatool (approved):
LGTM

PR #132, opened 2024-03-07
remove enterprise starter SLAs and add note about n-1 version support

enriquegh:


amenne (approved):
This LGTM for Enterprise. Did we want to add Cody Pro SLAs anywhere? Not blocking this but let's chat about how to capture that

PR #138, opened 2024-03-11
update documentation for permissions (RBAC and Repository)

BolajiOlajide:
I noticed in the grouping for `dev/how-tos`, we group content for Access Control and Repository permissions together, this PR separates that.

Also @bobheadxi suggested renaming the title of the Access Control docs for adding permissions so it's clear it's about RBAC permissions and not repository permissions.


MaedahBatool (approved):
LGTM 👍

PR #139, opened 2024-03-11
Add rendering for content type search results

enriquegh:
Adds search result styling for "content" type results which are slightly different than the title-based results


MaedahBatool (approved):
LGTM, awesome. 👍

PR #140, opened 2024-03-12
Clarify steps for Bitbucket Server repo permissions

eseliger:
I ran through this playbook and found two issues:

- When the username doesn't match the user used for the App Link, things break in weird ways.
- More recent versions of openssl don't generate PKCS1 keys anymore, which will fail saving the site configuration. This command makes it easier to run through the script.

## Test plan

Verified the docs by hand.


pjlast (approved):

MaedahBatool (approved):
LGTM thank you 👍

PR #143, opened 2024-03-12
Update messaging on how to resubscribe from a cancelled plan

enriquegh:
Based on a conversation with @chrsmith. Once the "Resume Subscription" button actually exists we can change this back.


chrsmith (approved):
LGTM.

/cc https://github.com/sourcegraph/accounts.sourcegraph.com/issues/433, which is tracking adding the ability to "reactivate" a canceled subscription.

PR #155, opened 2024-03-13
Fix Table of Contents errors

ahmadbilaldev:
This PR fixes TOC errors. They were caused by false headings being generated because of `#` inside code blocks. We are now ignoring all content inside code blocks while generating headings for TOC.

@MaedahBatool Please test if this solves the issues.


MaedahBatool (approved):
LGTM 👍

PR #160, opened 2024-03-15
Add tutorials page to navigation list

enriquegh:
This adds the tutorials page to the navigation list under "Help & Support"


enriquegh:
@shawnteeharris based on what we talked about.

This should also fix Algolia search since the crawler will now be able to find this page and will rank it appropriately.

MaedahBatool (approved):
As a quick fix, let's place tutorials here. In the long run we need to place it more strategically considering the overall information architecture. 

Approving the PR to unblock you. :)

PR #171, opened 2024-03-19
search jobs: default on

stefanhengl:
Search Jobs is on by default since 5.3.0. Reading the documentation, this is not obvious and customers have been confused about this. This little update should make it more obvious that Search Jobs works out of the box if the bundled blobstore is used.


jtibshirani (approved):
Makes sense!

MaedahBatool (approved):

PR #192, opened 2024-03-22
Support initiating search from hash url

mmanela:
I like to create browser shortcuts to quickly search sites. For the handbook you can do that with this url 
```
https://handbook.sourcegraph.com/#stq=test
```

However, for the docs site, it does not support this (as far as I can tell). But with this change you can do this
```
http://localhost:3000/#q=test
```

This PR adds support for this by having it look for the `#q=` hash param and using that to set the initial query. 

To support this I had to make a couple changes

1. In Search.tsx I extract the hash param if it is present and set the `initialQuery` prop in the `<DocSearch>` component (which was not passed in before)
2. I updated `<DocSearch>`  to automatically open the search dialog if the initialQuery prop is present and I clear the query when the dialog is closed. 

image

### DIFF
--- a/src/components/search/Search.tsx
+++ b/src/components/search/Search.tsx
@@ -8,7 +8,7 @@ import './docsearch/docsearch.css';
 const getInitialQuery = () => {
 	if (typeof window !== 'undefined' && window?.location?.href) {
 		const url = new URL(window.location.href);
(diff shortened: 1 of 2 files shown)


enriquegh (approved):
LGTM

MaedahBatool (approved):

PR #194, opened 2024-03-22
Add instructions to add images and binaries to docs readme

davidmedina-contact:
Add a block explaining how to add images


MaedahBatool (approved):
Perfect 👍

PR #229, opened 2024-04-03
Update feature parity matrix

taylorsperry:
Add updates for dynamic multi-line completion, edit command, ask Cody to fix command, and test command (inline vs chat)


MaedahBatool (approved):
LGTM

PR #235, opened 2024-04-04
Adding new anthropic claude 3 models family to supported models for enterprise customers

arafatkatze:
Adds claude 3 models to the list of supported models for enterprise customers and adds a green tick instead of coming soon. 

To be merged after https://github.com/sourcegraph/sourcegraph/pull/61601 and running the SQL query update


MaedahBatool (approved):
TY, I was being told it will go live tomorrow with the SG release. But thanks for your contribution. :)

PR #260, opened 2024-04-18
Update supported-models.mdx

Kynlos:
Updated model list to include Mistral 8x22b


MaedahBatool (approved):
Thank you, approved. 👍

PR #266, opened 2024-04-19
Updated Token Limits 

chillatom:
Syncs token limits with the latest updates


MaedahBatool (approved):
Hey @chillatom added some more suggestions to the PR via direct commit. :) Rest looks good let's merge it. 👍

PR #270, opened 2024-04-23
Add Cody Ignore pings

taras-yemets:
1. Docs update on new pings data:
- RFC
- https://github.com/sourcegraph/sourcegraph/pull/62080

2. Add missing data on Cody providers pings. It existed on the old docs site, data is still reported via pings, so it should probably be mentioned on the pings page. Feel free to push back on this if we intentionally removed this data.


MaedahBatool (approved):
LGTM, let's ship it.

PR #280, opened 2024-04-26
docs: explain language insights and config mechanisms

bahrmichael:
This PR updates the documentation as a follow-up to https://github.com/sourcegraph/sourcegraph/pull/62011 where we introduced new environment variables and improved the performance of language stats insights.

### DIFF
--- a/docs/code_insights/explanations/administration_and_security_of_code_insights.mdx
+++ b/docs/code_insights/explanations/administration_and_security_of_code_insights.mdx
@@ -59,3 +59,39 @@ The following setting(s) apply to adding new data to a previously backfilled Cod
 
 The following setting(s) apply to both backfilling data and adding new data
 - `insights.query.worker.rateLimit` - Maximum number of Code Insights queries initiated per second on a worker node.
+
+## Language Stats Performance Configuration (>= May 2024 Release)
+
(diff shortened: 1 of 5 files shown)


docs/code_insights/explanations/administration_and_security_of_code_insights.mdx
```diff
@@ -59,3 +59,39 @@ The following setting(s) apply to adding new data to a previously backfilled Cod
...
+
+### Language Stats Timeout
+
+The timeout in minutes for language stats queries is configurable through the `GET_INVENTORY_TIMEOUT` environment variable ([#62011](https://github.com/sourcegraph/sourcegraph/pull/62011)).
+
+Example:
+
+```
+GET_INVENTORY_TIMEOUT=5
+```
```
camdencheek, line 97: I'm not positive we want to be advertising these env vars. Generally, env vars are so we can have escape hatches in case of emergency, but we don't really want customers to be messing with them without help from support. I think it's probably fine to leave this section with the "reach out to Sourcegraph support"

camdencheek (approved):

docs/code_insights/explanations/administration_and_security_of_code_insights.mdx
bahrmichael: Where would we document this for our support folks?

PR #282, opened 2024-04-29
Add entry explaining external-accounts-modal feature flag to docs

pjlast:
To be merged with https://github.com/sourcegraph/sourcegraph/pull/61721


docs/admin/auth/index.mdx
```diff
@@ -568,3 +568,9 @@ Usernames from authentication providers are normalized before being used in Sour
 For example, a user whose external username (according the authentication provider) is `alice_smith@example.com` would have the Sourcegraph username `alice-smith`.
 
 If multiple accounts normalize into the same username, only the first user account is created. Other users won't be able to sign in. This is a rare occurrence; contact support if this is a blocker.
+
+## Remind users to connect external accounts
+
+When authorization is activated, users must link their external code host accounts with their Sourcegraph accounts to access private repositories.
+To ensure users are aware of any unlinked accounts, enable the external-accounts-modal feature flag.
```
mmanela, line 575: Do we want this to be an opt-in? Or should this be the default behavior BUT we let people opt-out?

eseliger: We talked about this in todays 1-1, we want to make this opt-in for the next release and gather feedback from the customer we're working with, and then if the feature works and solves their use-case we want to roll it out widely.

docs/admin/auth/index.mdx
```diff
@@ -568,3 +568,9 @@ Usernames from authentication providers are normalized before being used in Sour
...
+This prompts users to connect any missing external accounts upon visiting Sourcegraph.
```
eseliger, line 576: Maybe a screenshot of how this looks?

eseliger (approved):

PR #283, opened 2024-04-29
Add docs for sub-repo permissions

pjlast:
As part of https://github.com/sourcegraph/sourcegraph/issues/62105

This PR adds an entry to the explicit permissions API docs about sub-repo permissions and how to use them.

### DIFF
--- a/docs/admin/permissions/api.mdx
+++ b/docs/admin/permissions/api.mdx
@@ -145,6 +145,8 @@ mutation {
 }
 ```
 
+The default permission if a path is not specified is to deny access.
+
 ### Listing a user's authorized repositories
 
 You may query the set of repositories visible to a particular user with the `authorizedUserRepositories` [GraphQL API](/api/graphql) query, which accepts a `username` or `email` parameter to specify the user:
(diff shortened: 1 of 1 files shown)


docs/admin/permissions/api.mdx
```diff
@@ -95,6 +95,56 @@ Now, only the users specified in the `userPermissions` parameter will be allowed
...
+      }
+    ]
+  ) {
+    alwaysNil
+  }
+}
+```
+
+This query denies the user `alice` access to all files in the repository, except for the file `README.md`.
+The paths are specified in the [glob syntax](https://en.wikipedia.org/wiki/Glob_(programming)), and a `-` prefix indicates that the path should be denied.
```
eseliger, line 129: I believe `/cmd/` would also deny all paths, without the need to glob, is that correct?

Also, when a file matches none of the rules, does that mean deny or grant?

docs/admin/permissions/api.mdx
```diff
@@ -95,6 +95,56 @@ Now, only the users specified in the `userPermissions` parameter will be allowed
...
+Permissions are applied in the order they are specified. If we were to want to provide `alice` access to all files except for the `README.md` file, we could specify the following:
```
eseliger, line 130: How do I unset subrepo permissions? 

Can I also list active subrepo permissions, to diff what's already been ingested?

eseliger (approved):

MaedahBatool (approved):
LGTM :)

PR #286, opened 2024-04-30
Explain `internal` repositoryQuery

pjlast:
Doc update for https://github.com/sourcegraph/sourcegraph/pull/62266

### DIFF
--- a/docs/admin/external_service/github.mdx
+++ b/docs/admin/external_service/github.mdx
@@ -494,6 +494,8 @@ GitHub connections support the following configuration options, which are specif
 	//
 	// - `public` mirrors all public repositories for GitHub Enterprise and is the equivalent of `none` for GitHub
 	//
+	// - `internal` mirrors all internal repositories for GitHub Enterprise and is the equivalent of `none` for GitHub
+	//
 	// - `affiliated` mirrors all repositories affiliated with the configured token's user:
 	// 	- Private repositories with read access
 	// 	- Public repositories owned by the user or their orgs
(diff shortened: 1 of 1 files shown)


docs/admin/external_service/github.mdx
```diff
@@ -538,7 +538,7 @@ Sourcegraph displays search results from the default branch of a repository when
 ## Troubleshooting
 
 ### Hitting GitHub Search API rate limit with repositoryQuery
-When Sourcegraph syncs repositories configured  via `repositoryQuery`, it consumes GitHub API search rate limit, which is lower than the normal rate limit. The `affiliated`, `public` and `none` special values, however, trigger normal API requests instead of search API requests.
+When Sourcegraph syncs repositories configured  via `repositoryQuery`, it consumes GitHub API search rate limit, which is lower than the normal rate limit. The `affiliated`, `public`, and `none` special values, however, trigger normal API requests instead of search API requests. `internal` is also a special value that uses the GitHub Search API to list all internal repositories.
```
mmanela, line 541: Would it be useful to show a couple example code-host configs demonstrating these special values?

eseliger (approved):

docs/admin/external_service/github.mdx
MaedahBatool: +1 to @matt's suggestion. If we can add one it will be great. :)

pjlast: So there is the configuration section right above this that has all of the options for a code host connection: https://sourcegraph.com/docs/admin/external_service/github#configuration

And then also, right below this there's already an example of where `"repositoryQuery": ["affiliated"]` is used. 

I'm not sure if it makes sense to add another example here as well, since this trouble-shooting bit is regarding rate limits that are hit. Just added the bit about `internal` also using search quota here since it would also end up running into the rate limit issue.

mmanela (approved):

PR #294, opened 2024-05-03
shorter/simpler site description

sqs:
- "Delve" is a word that people have recently jokingly started to associate with AI-written content, so I'd like to avoid that word.
- It seems better to simplify the site description instead of using it to pitch Sourcegraph's value prop. The ideal would be to make each page's description and OpenGraph description describe the page rather than the overall site.


MaedahBatool (approved):
Thank you for the edits.

PR #297, opened 2024-05-06
Update token-limits of claude 3 sonnet and opus

enriquegh:
User-defined window for Opus and Sonnet is mentioned in this same doc as 30k but was added as 200k.
The number is mentioned as 30k here too:
https://sourcegraph.com/blog/cody-vscode-1-14-0-release#bigger-and-better-context-windows-to-improve-chat


enriquegh:
@chillatom Added you as reviewer since you were original author. Got a Cody Pro user asked about this and based on the info available I think its a typo but let me know if this is not true.

MaedahBatool (approved):
LGTM :)

PR #319, opened 2024-05-11
Update cloud docs

andreeleuterio:
These docs were misleading. We can support individual keys per-customer but it's a premium offering, not the default setup. I think adding that we may also support it only adds confusion.

## Pull Request approval


MaedahBatool (approved):
Thank you

PR #325, opened 2024-05-15
Remove content layout shifts (CLS) from Cody for VS Code install page

toolmantim:
Adds aspect ratios to the image and videos to prevent [CLS](https://web.dev/articles/optimize-cls), so we can reliably link to a section of the page without it scrolling away from it as elements load.

To generate the image tags I used:

```zsh
images=(
	https://storage.googleapis.com/abc.png
	https://storage.googleapis.com/def.png
);

for image in "${images[@]}"; do
	curl -O $image
	filename=$image:t
	aspect_ratio=$(file "$filename" | grep -Eo "[[:digit:]]+ *x *[[:digit:]]+" | sed -e 's/ x / \/ /')

	echo "<img"
	echo "	src=\"$image\""
	echo "	alt=\"\""
	echo "	style={{ aspectRatio: '$aspect_ratio' }}"
	echo "	/>"
done
```

### DIFF
--- a/docs/cody/clients/install-vscode.mdx
+++ b/docs/cody/clients/install-vscode.mdx
@@ -22,7 +22,11 @@ Follow these steps to install the Cody AI extension for VS Code:
 - Type **Cody AI** in the search bar and click the **Install** button
 - After installing, you may be prompted to reload VS Code to activate the extension
 
-![install-vscode-extension](https://storage.googleapis.com/sourcegraph-assets/Docs/install-cody-vscode.png)
+<img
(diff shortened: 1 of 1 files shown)


MaedahBatool:
@toolmantim, we can proceed with this as a temporary fix, but I can start working on a separate component to monitor and optimize the aspect ratio and CLS issue via Next Image.

MaedahBatool (approved):

toolmantim:
Nice! Great idea @MaedahBatool. I tried to add next/image quickly, but it didn't work so just opted for this quick fix.

Be cool to move off the Google Cloud storage bucket, so you can just have next/image calc the dimensions for you.

PR #382, opened 2024-06-03
Remove the deprecated Claude models from the list of supported models

ykdojo:
Source: https://github.com/sourcegraph/cody/blob/main/lib/shared/src/models/dotcom.ts


MaedahBatool (approved):
TY :)

PR #396, opened 2024-06-08
clarify docs for supported models and model providers

sqs:
This makes this section a full reference of the ways Cody Enterprise can use models and model providers, instead of just being about 3rd-party options. It's also not clear what "3rd-party" means, since arguably Anthropic and OpenAI models are 3rd-party even when used via Cody Gateway.

Needs to be merged before https://github.com/sourcegraph/cody/pull/4492


valerybugakov (approved):
LGTM in combination with https://github.com/sourcegraph/cody/pull/4492

PR #397, opened 2024-06-09
remove mentions of to-be-removed separate Cody natural-language search quickpick UI (prefer using it via chat instead)

sqs:
See https://github.com/sourcegraph/cody/pull/4506.

This capability is NOT being removed, but it's very rarely used (and only by non-enterprise users) separately. It is almost always used as part of Cody chat, which performs a natural-language search and shows the results in every chat in the context row.


MaedahBatool (approved):
Sounds good. Let's ship it then. :)

PR #413, opened 2024-06-11
Revert "Fixed logo path issue"

MaedahBatool:
Reverts sourcegraph/docs#412


MaedahBatool:
@zahidalidev I have reverted the PR as it was not working in production. 
CleanShot 2024-06-11 at 12 37 34@2x

The logo did break here as shown in the image. We might need to revisit this again.

PR #420, opened 2024-06-12
Revert "Update supported-models.mdx"

MaedahBatool:
Reverts sourcegraph/docs#419


aramaraju:
Confirmed with @abeatrix that Gemini works on Cody Enterprise, so my PR is actually valid 

Here's a screenshot on sg02 :

image



CleanShot 2024-06-13 at 11 27 01@2x

PR #428, opened 2024-06-13
Updating Tailwind font families

traceyljohnson:
Updated Tailwind font families to use Inter for both Sans and Display. Mono font is not Roboto Mono.


MaedahBatool (approved):
LGTM :)

PR #437, opened 2024-06-18
Prefer `brew install` on macOS as Homebrew is more ergonomic

qaisjp:
Per the tin: most macOS developers will have Homebrew, and I think they'll prefer installing through Homebrew instead of an unmanaged binary sitting on disk.


MaedahBatool:
@sourcegraph/stampers if anyone can you please help me review it? 🙏

camdencheek (approved):
LGTM, thanks!

unknwon (approved):

PR #458, opened 2024-06-26
Fixing link to Google paper. Double slash misplaced.

gl-yziquel:
The double slash in the link was misplaced. Correct url to research paper is https://static.googleusercontent.com/media/research.google.com//en/pubs/archive/43835.pdf


gl-yziquel:
Wrong. Nothing to change. Sorry for the PR.

PR #468, opened 2024-07-01
Add wording to `smartContextWindow` docs so the reader knows that the…

peterguy:
… setting is in the `completions` section


MaedahBatool (approved):
LGTM thanks :)

PR #487, opened 2024-07-11
search: update Stream API doc page

stefanhengl:
This documents all supported query parameters and also updates the curl example.
I also removed the "The Stream API is still evolving" banner.

### DIFF
--- a/docs/api/stream_api/index.mdx
+++ b/docs/api/stream_api/index.mdx
@@ -23,7 +23,8 @@ curl --header "Accept: text/event-stream" \
      --data-urlencode "t=<pattern type>" \
      --data-urlencode "max-line-len=<maximum line length>" \
      --data-urlencode "cm=<enable chunk matches>" \
-     --data-urlencode "display=<display limit>"
+     --data-urlencode "display=<display limit>" \
+     --data-urlencode "cl=<context lines>"
 ```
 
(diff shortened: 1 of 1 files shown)


jtibshirani:
This makes sense to me, but I'm wondering if there are certain parameters or design choices that we're not yet comfortable with? I'd be curious to hear from @keegancsmith and @camdencheek !

camdencheek (approved):
This is great -- thank you!

Unrelated to documenting it, but should we make chunk matches default true? I don't think we need to remove line matches (it's nearly zero maintenance burden), but given that we are only developing against chunk matches and some options (context lines) only work with chunk matches, I think it might make sense to encourage that as the default way to use the API.

jtibshirani (approved):
Ah, I thought that by removing the banner we were also declaring it "GA" and guaranteeing we won't make backwards incompatible changes. So that's why I was seeking broader input. Now I see this is likely not your intention?

For the docs updates themselves, +1 they look great!

stefanhengl:
> Ah, I thought that by removing the banner we were also declaring it "GA" and guaranteeing we won't make backwards incompatible changes. 

That was my intention :-). But ... defaulting to chunk matches makes sense to me, too. We could change the default with the next release and announce it in the changelog.

PR #494, opened 2024-07-13
Fix the searching from the url experience

mmanela:
The ability to search from a url added [here](https://github.com/sourcegraph/docs/pull/192)  became flakey / inconsistent. You would search using the hash string (`https://sourcegraph.com/docs#q=thingy`) and it would not pop open the search box until you gave focus to the window.

I debugged and could not figure out why this was happening exactly besides the conditional rendering of the portal modal dialog. So, to work around that, I always render the portal DOM and let the modal decide when it is visible or not. I am not sure if this will cause issues so needs some testing.


enriquegh (approved):

MaedahBatool (approved):
LGTM :)

PR #514, opened 2024-07-22
Reduce customer confusion by removing installation links to old releases

marcleblanc2:
Customers get confused about our position on how long they should stay on old versions, or which version they should install, when they see mixed signals in our docs and public handbook. Removing the links to install prior versions helps clarify that the paved path is only to install the latest release.


DaedalusG (approved):

DaedalusG:
Discussed briefly with team, don't see any problems with this since latest is what we'll want fresh installs to be, and upgrades can consult other docs 👍🏼

PR #516, opened 2024-07-23
Add docs for VS Code v1.28

MaedahBatool:
This PR adds changes to docs regarding the new experience with the unified Cody panel with Sourcegraph VS Code July release.

### DIFF
--- a/docs/cody/capabilities/chat.mdx
+++ b/docs/cody/capabilities/chat.mdx
@@ -37,9 +37,9 @@ For example, ask Cody "What does this file do?"
 
 Cody will take a few seconds to process your question, providing contextual information about the files it reads and generating the answer.
 
-        <video width="1920" height="1080" loop playsInline controls style={{ width: '100%', height: 'auto' }}>
-                    <source src="https://storage.googleapis.com/sourcegraph-assets/Docs/Media/ask-cody-question.mp4" type="video/mp4" />
-                    </video>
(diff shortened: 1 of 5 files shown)


docs/cody/capabilities/chat.mdx
```diff
@@ -37,9 +37,9 @@ For example, ask Cody "What does this file do?"
 
 Cody will take a few seconds to process your question, providing contextual information about the files it reads and generating the answer.
 
-        <video width="1920" height="1080" loop playsInline controls style={{ width: '100%', height: 'auto' }}>
-                    <source src="https://storage.googleapis.com/sourcegraph-assets/Docs/Media/ask-cody-question.mp4" type="video/mp4" />
-                    </video>
+<video width="1920" height="1080" loop playsInline controls style={{ width: '100%', height: 'auto', aspectRatio: '1920 / 1080' }}>
+  <source src="https://storage.googleapis.com/sourcegraph-assets/Docs/Media/chat-interface-0724.mp4" type="video/mp4" />
```
kalanchan, line 41: can we make a note that this is only available for PLG for now? Enterprise will come on August 7. 

and then follow up to remove this disclaimer after 😇

docs/cody/clients/install-vscode.mdx
```diff
@@ -130,9 +118,9 @@ function bubbleSort(array){
 
 ## Chat
```
kalanchan, line 119: same comment as above, a disclaimer saying this is only available for enterprise would be great :)

kalanchan (approved):
thanks Maedah! left a comment about adding a disclaimer and that should be good

PR #540, opened 2024-08-01
Update documentation to describe new format for LLM configuration

chrsmith:
In the August release, we will be introducing a new and much more flexible way for a Sourcegraph administrator to configure the LLM models used by Cody.

Previously the documentation for Cody Enterprise was found on the `clients/enable-cody-enterprise.mdx` ("Cody > Installation > Cody Enterprise" in the nav).

This PR moves all of the LLM-specific documentation into a new file, `clients/model-configuration.mdx` and a new "Model Configuration" nav element.

This new page then walks through the specifics of the site configuration options, and includes the previous documentation in a `# Older style "Completions" Configuration` section.

Let me know what you think.

- Is this sufficiently clear?
- Are there any scenarios that require more elaboration? (We can easily devote a dozen pages to examples of various permutations of configuration, but I wanted to err on the side of being concise. WDYT?)
- Is the call out to "you need to file a support ticket to make this change if you are a Sourcegraph cloud customer" make sense?

## Pull Request approval


docs/cody/capabilities/supported-models.mdx
```diff
@@ -4,6 +4,8 @@
 
 Cody supports a variety of cutting-edge large language models for use in Chat and Commands, allowing you to select the best model for your use case.
 
+<Callout type="note">Newer versions of Sourcegraph Enterprise can be configured to use newer models as they are introduced. See [Model Configuration](/cody/clients/model-configuration) for more information.</Callout>
+
```
aramaraju, line 8:
```suggestion
<Callout type="note">Newer versions of Sourcegraph Enterprise (starting from [Insert Release Version for Aug 7th] can be configured to use newer models as they are introduced. See [Model Configuration](/cody/clients/model-configuration) for more information.</Callout>
```

aramaraju: We need to fill out the release number when we know what it is going to be

aramaraju: Is it 5.5.4+?

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
+# LLM Model Configuration
+
+<p className="subtitle">This guide will walk you through the steps to customize the LLM models available from your Sourcegraph Enterprise instance.</p>
+
+<Callout type="note">For Sourcegraph Cloud customers, configuring the available LLM models requires contacting Sourcegraph's tech support.</Callout>
```
aramaraju, line 5:
```suggestion
<Callout type="note">For Sourcegraph Cloud customers, configuring the available LLM models requires contacting your Sourcegraph account team representative</Callout>

File: docs/cody/clients/model-configuration.mdx [hunk 3]
```diff
@@ -0,0 +1,520 @@
...
+
+Starting with Sourcegraph 5.5.4+ Sourcegraph Enterprise supports the ability to choose between different LLM models, allowing developers to use the best model for their current needs.
```
aramaraju, line 7:
```suggestion
Starting with Sourcegraph 5.5.4+ Sourcegraph Enterprise supports the ability to choose between different LLM models, allowing developers to use the best model for Cody Chat as needed.
```

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+
+This is accomplished exposing much more flexible configuration options for Cody when using Sourcegraph Enterprise. The newer style of configuration is described next. However, you can still use the [Older style "Completions" Configuration](#Older-style-Completions-Configuration).
```
aramaraju, line 9: This is okay for Aug 7th, but I suspect that we will want to have better 'customer facing' names in Sept

cc: @kelvinyap

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+      "allow": [
+        "anthropic::*",
+        "openai::*"
+      ],
+
+      // Do not include any models with the Model ID containing "turbo",
+      // or any from AcmeCo.
+      "deny": [
+        "*turbo*",
+        "acmeco::*"
```
aramaraju, line 75: Should we link this to the Supported Models Matrix, with the configuration names ?

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+    }
+  }
+}
+```
+
+## Default Models
+
+The `"modelConfiguration"` setting also contains a `"defaultModels"` field that allows you to specify the LLM model used depending on the situation. If no default is specified, or refers to a model that isn't found, it will silently fallback to a suitable alternative.
+
+You can leave these settings blank and have reasonable defaults be chosen for you, or you can express the models explicitly.
```
aramaraju, line 86:
```suggestion
 use reasonable defaults chosen for you by Sourcegraph
```

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+  ...
+```
+
+The format of these strings is a "Model Reference", which is a format for uniquely identifying each LLM model exposed from your Sourcegraph instance.
+
+## Advanced Configruation
+
+For most administrators, relying on the LLM models made available by Cody Gateway is sufficient. However, if even more customization is required, you can configure your own LLM providers and models.
+
+<Callout type="warning">Defining your own LLM providers and models is an advanced use-case and requires care to get the correct results. It also may bypass protections to ensure compatibiloity between your Sourcegraph instance and LLMs. If you need help contact your Sourcegraph account executive.</Callout>
```
aramaraju, line 106:
```suggestion
<Callout type="warning">Defining your own LLM providers and models is an advanced use-case and requires care to get the correct results. It also may bypass protections to ensure compatibility between your Sourcegraph instance and LLMs. If you need help contact your Sourcegraph account executive.</Callout>
```

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+```json
+"modelConfiguration": {
+  // Do not use any Sourcegraph-supplied models.
+  "sourcegraph": null,
+
+  // Introduce our own, custom provider "anthropic-direct".
+  "providerOverrides": [
+    {
+      "id": "anthropic-direct",
+      "displayName": "Anthropic models, sent directly to anthropic.com",
```
aramaraju, line 129: This display name shows up on the logs and only the logs, right?

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+      },
+
+      // The default model configuration provides defaults for all LLM
+      // models using this provider.
+      "defaultModelConfig": {
+        "capabilities": [
+          "chat",
+          "autocomplete"
+        ],
+        "contextWindow": {
```
aramaraju, line 145: @slimsag How does smartContextWindow come into all this configuration btw? Do I get it if I don't change the contextWindow on the models that are supported?

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+      "status": "stable"
+    },
+  ],
+  ...
+}
+```
+
+Most of the configuration fields are self-explanitory, such as labeling the model's category ("stable", "beta") or category ("accuracy", "speed"). The more important fields are described below:
+
+**modelRef**. Each model is given a unique identify, referred to as a model reference or "mref". This is a string of the form `${providerId}::${apiVersionId}::${modelId}`.
```
aramaraju, line 211:
```suggestion
**modelRef**. Each model is given a unique identifier, referred to as a model reference or "mref". This is a string of the form `${providerId}::${apiVersionId}::${modelId}`.
```

src/data/navigation.ts
```diff
@@ -343,6 +344,7 @@ const navigation_5_2: NavigationItem[] = [
               { title: "Cody for Neovim", href: "/cody/clients/install-neovim", },
               { title: "Cody for Web", href: "/cody/clients/cody-with-sourcegraph", },
               { title: "Cody for Enterprise", href: "/cody/clients/enable-cody-enterprise", },
+              { title: "LLM Model Configuration", href: "/cody/clients/model-configuration", },
```
aramaraju, line 347: @MaedahBatool As a meta point, the docs and layout in the "Enabling Cody Enterprise" page is getting very long. Let's start thinking about better organizing the data after Aug 7th :)

aramaraju (changes requested):
Left a few comments/suggestions. Thank you @chrsmith

docs/cody/capabilities/supported-models.mdx
MaedahBatool: +1 to @aramaraju

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+
+**contextWindow**. The context window is the number of "tokens" sent to an LLM. Either in the amount of contextual data sent in the LLM prompt (e.g. the question, relevant snippets, etc.) and the maximum size of the output allowed in the response. These values directly control factors such as the time it takes to respond to a prompt and the cost of the LLM request. And each LLM model or provider may have their own limits as well.
+
+**modelName**. The model _name_, is the value required by the LLM model's API provider. In this example, the `modelRef` defined the model's ID as `claude-3-sonnet` but the `modelName` was the more specific "claude-3-sonnet-20240229".
+
+**capabilities**. The capabilities of a model determine which situations the model can be used. For example, models only supported for "autocomplete" will not be available for Cody chats.
+
+# Older style "Completions" Configuration
```
MaedahBatool, line 223:
```suggestion
## Legacy Completions Configuration
```
Let's call it legacy instead of older. Also we should make it an H2.

src/data/navigation.ts
MaedahBatool: I agree with you 100% I have plans to revamp the entire Cody docs by Q3 with a better layout.

docs/cody/capabilities/supported-models.mdx
kalanchan: yeah I'll confirm with the release team, they are only able to bump minor versions every 2 months for now so the aug 7 release is not predictable

docs/cody/clients/model-configuration.mdx
chrsmith: Currently, I don't know if we show the display name _anywhere_. The provider _ID_ will show up in logs (since it is baked into the model references.)

The intent of adding a separate display name field was so that we could display this human-friendly text to users, e.g. when differentiating between various flavors of Claude in the Cody IDE extension.

docs/cody/clients/model-configuration.mdx
chrsmith: Currently, if you are using the new modelconfig version of site config, we set smartContext to "disabled". See https://github.com/sourcegraph/sourcegraph/blob/e5178a6bc0e0f2f5208f9dedb0c226661c881d98/cmd/frontend/internal/modelconfig/resolver.go#L94-L107.

I'm not sure what the ramifications of that is, since we are passing the more specialized context window sizes down to the cody client with the new model metadata. (So rather than _only_ using `[siteconfig]completions.maxChatTokens` we instead have a per-model value, defined in the configuration here.)

Does that all sound right? If using modelconfig should we default smartContext to "enabled"?

aramaraju: It's probably best to leave that enabled by default. The Client side logic will take care of whether the models are supported by it or not (can you confirm this @abeatrix ?)

docs/cody/clients/model-configuration.mdx
aramaraju: Got it - thanks for confirming Chris!

src/data/navigation.ts
aramaraju: I added a project for Merge to get this rolling :)

docs/cody/clients/model-configuration.mdx
abeatrix: @jamesmcnamara is probably the best person to answer this for the client side

jamesmcnamara: I am not familiar with the smart context window, but from what I can see I see most of the uses related to smart context window are for old style models and the few references I see to the models specific context window are don't intersect with smart context feature enablement (i.e. they don't seem to care if that flag is enabled or not they just get the models context window).

But I am not familiar with this issue so I could be wrong, there might be some sneaky call at the top of the tree that does actually care about that setting.

chrsmith:
Quick update: I'm currently putting out various brush fires to ensure the release goes out smoothly. It seems very likely that this won't be updated by the EOD today, and so it'll miss the train for the `sg-next-aug2024` docs release.

I'll update this to instead target `main`, and figure out how to publish these docs as soon as they are ready. (But perhaps after the August release goes out.)

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,520 @@
...
+```json
+"modelConfiguration": {
+  "sourcegraph": {
+    "modelFilters": {
+      // Only allow "beta" and "stable" models.
+      // Not "experimental" or "deprecated".
+      "statusFilter": ["beta", "stable"],
+
+      // Allow any models provided by Anthropic or OpenAI.
+      "allow": [
```
aramaraju, line 66: @chrsmith YOU HAD IT RIGHT all along!! HAHA 😆

chrsmith:
I'll close this out and send out a new PR, rebased on top of `main`.

PR #546, opened 2024-08-06
Docs for Prompts and Prompt Library

MaedahBatool:
Adds docs for the upcoming Prompts and the Prompt Library.

- https://linear.app/sourcegraph/issue/PROD-125/create-docs-for-prompts-and-prompt-library

### DIFF
--- a/docs/cody/capabilities/commands.mdx
+++ b/docs/cody/capabilities/commands.mdx
@@ -30,7 +30,7 @@ Completing this process will do the following:
 - If it's not a draft, it will be available to other users in your organization
 - The prompt will start appearing via your connected Sourcegraph instance inside Cody's VS Code extension and the Sourcegraph Web UI
 
-#### VIDEO DEMO TO CREATE A Prompt
+![prompts-in-vscode](https://storage.googleapis.com/sourcegraph-assets/Docs/prompts-ui-vscode-082024.png)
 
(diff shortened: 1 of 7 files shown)


MaedahBatool:
@chenkc805, can you please review the content? I still need to add the images and videos, though. Waiting for Quinn to push recent UI changes.

chenkc805:
Thanks Maedah! I'm currently commuting so can't give in depth comments, but at a high level here's some feedback:
- Let's make Prompts the primary, and Commands secondary. 
- Instances of "custom commands" should tie the idea that "custom commands" is being replaced by Prompts
- Commands and Custom Commands are going to be replaced by Prompts and Prompt Library in the coming future
- For the Cody quick start, can we do the following: 
  - add generate an inline edit
  - add accept an autocomplete
  - add "@"mention a repo or file in a chat to give specific context
  - add create a prompt in your prompt library
  - take out generate unit tests, as this functionality doesn't perform well.

MaedahBatool:
@chenkc805 thanks for the feedback. I have addressed all your feedback except for your suggestions on revamping the Quickstart guide. I totally agree with your proposal. I am taking it as a follow-up task for next week. Let's ship the docs for Prompts and Prompt library to unblock the release work and I can ship the improved Quickstart next week. :)

PR #555, opened 2024-08-08
Add documentation for the newer modelconfig settings

chrsmith:
This is the same content as in https://github.com/sourcegraph/docs/pull/540, just with some of the requested changes applied.

---

In the August release, we will be introducing a new and much more flexible way for a Sourcegraph administrator to configure the LLM models used by Cody.

Previously the documentation for Cody Enterprise was found on the `clients/enable-cody-enterprise.mdx` ("Cody > Installation > Cody Enterprise" in the nav).

This PR moves all of the LLM-specific documentation into a new file, `clients/model-configuration.mdx` and a new "Model Configuration" nav element.

This new page then walks through the specifics of the site configuration options, and includes the previous documentation in a `# Older style "Completions" Configuration` section.

Let me know what you think.

- Is this sufficiently clear?
- Are there any scenarios that require more elaboration? (We can easily devote a dozen pages to examples of various permutations of configuration, but I wanted to err on the side of being concise. WDYT?)
- Is the call out to "you need to file a support ticket to make this change if you are a Sourcegraph cloud customer" make sense?

## Pull Request approval


docs/cody/capabilities/supported-models.mdx
```diff
@@ -4,6 +4,8 @@
 
 Cody supports a variety of cutting-edge large language models for use in Chat and Commands, allowing you to select the best model for your use case.
 
+<Callout type="note">Newer versions of Sourcegraph Enterprise (starting from v5.6) can be configured to use newer models as they are introduced. See [Model Configuration](/cody/clients/model-configuration) for more information.</Callout>
```
aramaraju, line 7: @chrsmith 
Doesn't this statement imply that the new models will automatically be pulled? I thought this wasn't done for 5.6? 

(Also worth mentioning this is only for Cody Gateway customers, right?)

docs/cody/clients/model-configuration.mdx
```diff
@@ -0,0 +1,518 @@
+# LLM Model Configuration
+
+<p className="subtitle">This guide will walk you through the steps to customize the LLM models available from your Sourcegraph Enterprise instance.</p>
+
+<Callout type="note">For Sourcegraph Cloud customers, configuring the available LLM models requires contacting your Sourcegraph account team representative</Callout>
+
+Starting with Sourcegraph 5.5.4+ Sourcegraph Enterprise supports the ability to choose between different LLM models, allowing developers to use the best model for Cody Chat as needed.
```
aramaraju, line 7: I believe the release on Aug 7th was 5.6.0, right?

aramaraju (changes requested):
Let's clarify the version numbers, and then we are good here

docs/cody/clients/model-configuration.mdx
chrsmith: Yes, it was. Good catch.

docs/cody/capabilities/supported-models.mdx
chrsmith: I was kinda sorta intentionally trying to be vague here, just to avoid getting into "right now they don't" and then needing to update it later to say "oh yeah, now it does".

So "as newer models are introduced" is really "[you can update your config] as newer models are introduced... but that's a pretty weak excuse and I'll update this to be clearer.

What do you think about:

"Starting with sourcegraph vX yada yada it will be even easier to add support for new models and providers, see [Model Configuration] for more information."

aramaraju: I like that rewrite 👍 Thank you 🙇

aramaraju (approved):
🙇

PR #565, opened 2024-08-14
various typos and link fix

loujar:
Some cleanup for this new doc page. I have some general feedback on the design of this doc outside of these changes:
- This doc seems to be written primarily for site admins that are using Cody Gateway, and doesn't address the BYOK ("advanced") configuration options until later in the doc. I think linking to the [Advanced Configuration](https://sourcegraph.com/docs/cody/clients/model-configuration#advanced-configruation) section earlier in the introductory paragraphs would help make this doc easier to navigate for BYOK admins.
- It's unclear whether how [Default Models](https://sourcegraph.com/docs/cody/clients/model-configuration#default-models) works on BYOK instances, or instances where `"sourcegraph": null` is configured. Specifically, how are "suitable alternatives" determined in these cases?
- This has probably already come up, but it seems odd / contradictory that we still document the (deprecated) `completions` configuration in the [site config page](https://sourcegraph.com/docs/admin/config/site_config) but don't document the (new) `modelConfiguration` section there.

### DIFF
--- a/docs/cody/clients/model-configuration.mdx
+++ b/docs/cody/clients/model-configuration.mdx
@@ -6,11 +6,11 @@
 
 Sourcegraph v5.6.0 or later supports the ability to choose between different LLM models, allowing developers to use the best model for Cody Chat as needed.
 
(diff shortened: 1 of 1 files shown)


MaedahBatool (approved):
LGTM, thanks :)

PR #567, opened 2024-08-14
Abstract older versions docs rerouting

noah-berman:
Older versions of the docs get rerouted to the site in the style of `https://${MAJOR}.${MINOR}.sourcegraph.com/docs`, for which we use the NextJS built-in redirect. This PR abstracts all older versions (those other than the NextJS-hydrated `config.DOCS_LATEST_VERSION`, preventing versions previously unlisted from being excluded from the pass-through (and resulting in broken docs links on our site).

## Testing

Run locally and tested with the various `source` permutation values.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -23,45 +23,45 @@ const nextConfig = {
 				permanent: false
 			},
 			// // Redirect for docs for version 5.5 hosted at https://5.5.sourcegraph.com/
-			// {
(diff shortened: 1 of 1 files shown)


MaedahBatool:
Thanks @noah-berman, for your help, but while testing, the redirects failed. So closing this PR in response to this one: https://github.com/sourcegraph/docs/pull/570/files

PR #574, opened 2024-08-15
Add Okta SCIM private provisioning callout

marcleblanc2:
## Pull Request approval

### DIFF
--- a/docs/admin/scim.mdx
+++ b/docs/admin/scim.mdx
@@ -6,6 +6,8 @@ SCIM (System for Cross-domain Identity Management) is a standard for provisionin
 
 Sourcegraph supports SCIM 2.0 for provisioning and de-provisioning _users_.
 
+> <Callout type="note">SCIM integration requires your SCIM provider to have network connectivity to your Sourcegraph instance. For example, using Okta's SCIM service with a private Sourcegraph instance, you will need to implement an [Okta privisioning agent](https://help.okta.com/en-us/content/topics/provisioning/opp/opp-main.htm) within a network that can connect to both your Sourcegraph instance and Okta's cloud services before continuing this setup.</Callout>
+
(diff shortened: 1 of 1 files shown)


gabtorre (approved):

docs/admin/scim.mdx
```diff
@@ -6,6 +6,8 @@ SCIM (System for Cross-domain Identity Management) is a standard for provisionin
 
 Sourcegraph supports SCIM 2.0 for provisioning and de-provisioning _users_.
 
+> <Callout type="note">SCIM integration requires your SCIM provider to have network connectivity to your Sourcegraph instance. For example, using Okta's SCIM service with a private Sourcegraph instance, you will need to implement an [Okta privisioning agent](https://help.okta.com/en-us/content/topics/provisioning/opp/opp-main.htm) within a network that can connect to both your Sourcegraph instance and Okta's cloud services before continuing this setup.</Callout>
```
loujar, line 9: > private Sourcegraph instance

This is maybe not the clearest way to articulate this. Technically we're talking about sourcegraph instances hosted in a private network without internet connectivity. One could argue that pretty much every sourcegraph instance is a "private instance" (with the exception of sourcegraph.com/search)

marcleblanc2: That's fair. We can give it some thought and see if there's a succinct way to provide the example. We could also remove the example entirely, as "SCIM integration requires your SCIM provider to have network connectivity to your Sourcegraph instance" may be clear enough on its own.

image

marcleblanc2: Sourcegraph Cloud customers (without IP allowlisting enabled) would already have network connectivity open from Okta for SCIM, and those with IP allowlisting enabled could add Okta's public IPs to the allowlist. 

For self-hosted customers, there's not a ton of them which have inbound connectivity open from the internet to their Sourcegraph instance (except maybe for webhooks from GitHub.com?), so they're fairly squarely in the scope for self-hosted Okta provisioners.

PR #620, opened 2024-09-04
Fix undefined suggested search results 

enriquegh:
Algolia stores the lvl0 hits as suggested searches. Our docsite currently only has 1 lvl0 (Documentation) causing the searches to be not useful.

As a quick fix, adding "Cody" and "Code Search" as initial suggestions fixes this. Long term, changing how our docs are getting scraped and getting the navbar headers to act as lvl0 is probably preferred.

Screenshot 2024-09-04 at 08 06 47


MaedahBatool (approved):
LGTM, TY :)

PR #638, opened 2024-09-09
Remove cody.enable step when getting verbose logs and use export logs command

enriquegh:
Since 1.16.0 we no longer need to enable cody debug it is enabled by default. This updates doc to remove this and also use the export logs command for easier log retrieval


morgangauth (approved):
lgtm!

PR #645, opened 2024-09-10
Update model-configuration.mdx

aramaraju:
Restructured the docs to make it more obvious that Model Configuration is in EAP

## Pull Request approval

### DIFF
--- a/docs/cody/clients/model-configuration.mdx
+++ b/docs/cody/clients/model-configuration.mdx
@@ -4,223 +4,14 @@
 
 <Callout type="note">For Sourcegraph Cloud customers, configuring the available LLM models requires contacting your Sourcegraph account team representative</Callout>
 
-Sourcegraph v5.6.0 or later supports the ability to choose between different LLM models, allowing developers to use the best model for Cody Chat as needed.
+Cody Enterprise can be configured using one of two methods:
+1. Legacy "Completions" Configuration
+2. Model Configuration (Early Access Program)
 
-This is accomplished exposing much more flexible configuration options for Cody when using Sourcegraph Enterprise. The newer style of configuration is described next. However, you can still use the [Older style "Completions" Configuration](#legacy-completions-configuration).
+<Callout type="warning">Configuring Cody via your Sourcegraph Enterprise instance via the Legacy "Completions" Configuration site configuration section is going to be deprecated soon. For now, it will be backwards compatible and supported. The Model Configuration method is in Early Access Program and only avaiable on Sourcegraph v5.6.0 or later </Callout>
 
-## Quickstart
+## Legacy "Completions" Configuration
 
-<Callout type="info">Before you start, please note that the model configuration is an early access program (EAP) and we are working to make the experience better.</Callout> 
-
-The simplest way to configure your Sourcegraph Enterprise would be to add the following configuration section to your instance's [site configuration](/admin/config/site_config):
-
-```json
-  ...
-  "modelConfiguration": {
-    "sourcegraph": {}
-  },
-  ...
-```
-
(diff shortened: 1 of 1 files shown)


docs/cody/clients/model-configuration.mdx
```diff
@@ -4,223 +4,14 @@
 
 <Callout type="note">For Sourcegraph Cloud customers, configuring the available LLM models requires contacting your Sourcegraph account team representative</Callout>
 
-Sourcegraph v5.6.0 or later supports the ability to choose between different LLM models, allowing developers to use the best model for Cody Chat as needed.
+Cody Enterprise can be configured using one of two methods:
+1. Legacy "Completions" Configuration
```
loujar, line 8: I personally feel we shouldn't refer to this as the legacy option if that is our recommendation. Is there a specific reason why we're using this language?

docs/cody/clients/model-configuration.mdx
```diff
@@ -4,223 +4,14 @@
...
+2. Model Configuration (Early Access Program)
 
-This is accomplished exposing much more flexible configuration options for Cody when using Sourcegraph Enterprise. The newer style of configuration is described next. However, you can still use the [Older style "Completions" Configuration](#legacy-completions-configuration).
+<Callout type="warning">Configuring Cody via your Sourcegraph Enterprise instance via the Legacy "Completions" Configuration site configuration section is going to be deprecated soon. For now, it will be backwards compatible and supported. The Model Configuration method is in Early Access Program and only avaiable on Sourcegraph v5.6.0 or later </Callout>
```
loujar, line 11: Again, I don't think we should refer to this option as deprecated as long as its our primary recommendation.

docs/cody/clients/model-configuration.mdx
```diff
@@ -515,4 +306,221 @@ To enable StarCoder go to **Site admin > Site configuration** (`/site-admin/conf
...
+
+**Server-side Configuration**
+
+The most important part of a provider's configuration is the `"serverSideConfig"` field. That defines how the LLM model's should be invoked, i.e. which external service or API will be called to serve LLM requests.
+
+In the example, the `"type"` field was `"anthropic"`. Meaning that any interactions using the `"anthropic"` provider would be sent directly to Anthropic, at the supplied `endpoint` URL using the given `accessToken`.
+
+However, Sourcegraph supports several different types of LLM API providers natively. The current set of supported LLM API providers is:
+
+| Provider type        | Description  |
```
loujar, line 469: Noting that we still do not document the correct options for `providerOverrides.[].id` anywhere in this doc. I think it would be good to include this information here in the same table, since I assume there is 1 to 1 mapping between `providerOverrides.[].id` and `providerOverrides.[]serverSideConfig.type`

docs/cody/clients/model-configuration.mdx
slimsag: Reiterating myself:

> I would expect the docs phrasing to read more like ‘Versions of Sourcegraph after Aug 7th feature a fancy new way to configure models and model providers, this is early-access, it might work for you but if you don’t want to be an early tester than you should use completions instead <link>’

docs/cody/clients/model-configuration.mdx
slimsag: Agreed

docs/cody/clients/model-configuration.mdx
slimsag: Agreed and described in more detail here the work we should be doing: https://linear.app/sourcegraph/issue/PRIME-526/document-accepted-values-for-provideroverrides%5B%5Did-in-new

loujar (approved):

PR #675, opened 2024-09-26
Add FAQ for using model of preference on Cody CLI

airamare01:
This update introduces an answer to a question where users would like to find out if they can use a model of choice to chat with Cody on CLI.

## Pull Request approval


Rhia2 (approved):
Thank you for including this boss

PR #676, opened 2024-09-26
Fix: Get Version string from the releaseRegistry

Chickensoupwithrice:
- **Hackathon attempt to get latest version into docs**
- **POC for Versions being updated from the ReleaseRegistry**
- **Remove Versions.tsx, it's no longer required**
- **Update all instances of "v5.6.2535" to be {CURRENT_VERSION}**
- **remove flake.nix whoops**

Our version strings are always out of date because we don't have any automation to update them. This PR resolves that problem by reaching out to the ReleaseRegistry at build time to get the latest version string. It then replaces that string in any code block using the PreCodeBlock component (which was previously broken and is now fixed). 
It also adds an exported const `CURRENT_VERSION_STRING` which can be used outside of Code Blocks and is similarly populated at build time. 

The final stage of this PR is automation that kicks off a build and releases the docs from the release pipeline itself so nobody has to go in and do that manually. 

## Pull Request approval


MaedahBatool:
Hey @Chickensoupwithrice the Vercel build is failing. Looks like a `type` issue. You can run `pnpm run build` to view build errors locally.

Chickensoupwithrice:
Yes `CURRENT_VERSION_STRING` currently doesn't work, I still need to fix that. Converted to draft to avoid confusion for the time being.

Chickensoupwithrice:
We still need automation to build the docs again on new release, which I think is a better solution than moving the current_version to client side

jdpleiness:
> We still need automation to build the docs again on new release, which I think is a better solution than moving the current_version to client side 

They get rebuilt during our release process when we merge the changelog anyhow, I believe. While I agree automation is the way to go, we can probably split that into another PR down the line. wdyt?

jdpleiness (approved):
🚀 🙏

src/components/MdxComponents.tsx
```diff
@@ -9,63 +9,63 @@ import { LinkCard, LinkCards } from './mdx/LinkCards';
...
-			const isCodeBlock = props.style?.display === 'grid';
-			// Only apply styles to inline code
-			const codeClasses = isCodeBlock
-				? 'bg-transparent'
-				: 'border font-medium bg-slate-100 dark:bg-slate-900 px-1 py-0.5 rounded-md border-slate-300 dark:border-slate-700';
+  return {
+    FeatureParity,
+    ResourceEstimator,
+    AWSOneClickLaunchForm,
+    Accordion,
```
DaedalusG, line 21: Indentations 🤦🏼‍♂️

Chickensoupwithrice:
> They get rebuilt during our release process when we merge the changelog anyhow, I believe. While I agree automation is the way to go, we can probably split that into another PR down the line

Yes absolutely.

src/components/MdxComponents.tsx
```diff
@@ -9,63 +9,63 @@ import { LinkCard, LinkCards } from './mdx/LinkCards';
...
+    ProductCards,
+    ProductCard: (props: React.ComponentProps<typeof ProductCard>) => (
+      <ProductCard {...props} version={version} />
+    ),
+    code: (props: any) => {
+      const isCodeBlock = props.style?.display === 'grid';
+      // Only apply styles to inline code
+      const codeClasses = isCodeBlock
+        ? 'bg-transparent'
+        : 'border font-medium bg-slate-100 dark:bg-slate-900 px-1 py-0.5 rounded-md border-slate-300 dark:border-slate-700';
```
Chickensoupwithrice, line 45: We can remove these lines now, they exist in the PreCode component

src/components/PreCodeBlock.tsx
```diff
@@ -5,23 +5,113 @@ interface PreProps extends DetailedHTMLProps<HTMLAttributes<HTMLPreElement>, HTM
...
+    React.Children.forEach(children, (child) => {
+      if (React.isValidElement(child) && child.props.children) {
+        codeContent += child.props.children
+      }
+    })
+  }
+
+  const version = await getLatestVersion()
+  const version_no_v = trimPrefix(version, "v")
+  codeContent = codeContent.replace(/{CURRENT_VERSION}/g, version)
```
DaedalusG, line 75: Nice so no special release just whenever the docsite builds this parent component it'll get the info from the release registry 🚀

DaedalusG (approved):

src/components/PreCodeBlock.tsx
```diff
@@ -5,23 +5,113 @@ interface PreProps extends DetailedHTMLProps<HTMLAttributes<HTMLPreElement>, HTM
...
+  let codeContent = props?.raw || ''
+
+  // If raw prop is not available, try to extract content from children
+  if (!codeContent && children) {
+    React.Children.forEach(children, (child) => {
+      if (React.isValidElement(child) && child.props.children) {
+        codeContent += child.props.children
+      }
+    })
+  }
```
Chickensoupwithrice, line 104: We only use `children` here

Chickensoupwithrice: (aka can remove this bit)

src/components/MdxComponents.tsx
```diff
@@ -9,63 +9,63 @@ import { LinkCard, LinkCards } from './mdx/LinkCards';
...
```
Chickensoupwithrice, line 44: This lives in the `PreCode` component now, can also be removed

src/components/MdxComponents.tsx
Chickensoupwithrice: I blame `vstls`

src/components/PreCodeBlock.tsx
Chickensoupwithrice: Yessir! We have 2 components, and 2 replacements in each of them. One for inline code blocks, and one for entire <pre> code blocks. Then there's one replacement for the version with the `v`, and one for without the `v`

DaedalusG:
This change is probably worth a note somewhere in the doc writing handbook, inquiring about that here: https://sourcegraph.slack.com/archives/C01DXLN3D0T/p1727460257629219

PR #680, opened 2024-10-01
Update index.mdx for Deployment Details and Recommendations

anorrish:
Updating the deployment landing page and including our recommendations.


loujar (approved):
Looks fine to me, but a few thoughts on the contents of the recommendations:
1. For Sourcegraph Cloud, we say that we handle "most" of the maintenance, etc. I would personally use the word "all" instead
2. For k8s helm: I would rephrase "... supported within many Kubernetes-capable environments" to "supported by all mainstream managed Kubernetes platforms"
3. For k8s helm, I don't quite follow why we refer to kustomize as a "more advanced solution" than helm. That might lead the user to think that kustomize offers greater customization than helm (which is sort of true, but we would want to enhance the templating ability of helm wherever that is the case and can easily address those changes quickly for customers when they come up)

PR #682, opened 2024-10-01
Search: clarify file filter description

jtibshirani:
This PR clarifies the description of the `file:regexp-pattern` search filter. The regexp is not anchored by default, and can match against any substring of the path.

Resolves this thread: https://sourcegraph.slack.com/archives/C05R619V4F8/p1727689383280839


jtibshirani:
For those who can't see the preview, here's how it looks now:
Screenshot 2024-10-01 at 12 00 00 PM

varungandhi-src (approved):
LGTM, thanks!

PR #685, opened 2024-10-02
Page layout for API docs

zahidalidev:
## Pull Request approval


src/components/Layout.tsx
```diff
@@ -88,23 +88,27 @@ function Header() {
 export function Layout({ children }: { children: React.ReactNode }) {
 	let pathname = usePathname();
 	let isHomePage = pathname === '/';
+	let isCodyApiPage = pathname === '/cody-api';
```
olafurpg, line 91: The OpenAPI spec is for "Sourcegraph API" and I think it's best that we update the page https://sourcegraph.com/docs/api to be the home for these docs. I will update the OpenAPI spec to include the contents from the current https://sourcegraph.com/docs/api/stream_api https://sourcegraph.com/docs/api/graphql pages

The end goal is that these API docs will cover all of our APIs, not just Cody.

olafurpg (commented):
Exciting! I'm very much looking forward to seeing these docs go live :D 

The select menu needs customization for dark theme

https://github.com/user-attachments/assets/eda3de8e-2c49-4726-a23f-1d43819d0287

We should enable text wrapping for the authorization component

CleanShot 2024-10-09 at 15 46 31@2x

CleanShot 2024-10-09 at 15 46 49@2x

(low-priority) The total number of languages here is a bit overwhelming, we may want to reduce the number of options if possible 
CleanShot 2024-10-09 at 15 47 14@2x

src/components/Layout.tsx
olafurpg: After more thinking, it's probably best to call this the "OpenAPI explorer" for "Sourcegraph APIs". We can treat this as a sibling page to the graphql and streaming search. The OpenAPI explorer will also document streaming search + graphql, but it's OK if there's some duplicate information. We can't remove the custom streaming search page or graphql page

olafurpg:
minor: the "export" option downloads a file named "document.yaml", it would be nice if it used the same name as the original spec file.

MaedahBatool:
Closing temporarily.

PR #744, opened 2024-11-01
Add fork:no and archived:no as search considerations

noah-berman:
Adds details around the implied fork:no and archived:no search results to search considerations page. This may need to be updated if/when these implied values are removed. 

See Linear comment [here](https://linear.app/sourcegraph/issue/SRCH-1305/morgan-stanley-repo-not-showing-up-in-search-results-unless-specified#comment-161dd719)


camdencheek (approved):
Thanks!

PR #772, opened 2024-11-05
Hackathon - Openctx

kukicado:
## Pull Request approval


chris-sev:
@kukicado @MaedahBatool  We have to double check if Web URLs are indeed turned on by default

They were for me, but could use confirmation

MaedahBatool:
> @kukicado @MaedahBatool We have to double check if Web URLs are indeed turned on by default
> 
> They were for me, but could use confirmation

@chris-sev I checked with a new account. Web URLs are enabled by default. I just made a minor fix. We are good to go with this one. Thanks

MaedahBatool (approved):

PR #795, opened 2024-11-12
Fix broken URL in quickstart.mdx

jaywgraves:
Found this bad URL when reviewing the docs


jdorfman:
Hi @jaywgraves we are purposely linking to our code search page. It loads for me, is it not loading for you?

jaywgraves:
it didn't when I first tried but now it does.  something weird on my end.   ok to reject.

PR #799, opened 2024-11-13
all: update links to sourcegraph blobs to public snapshot

keegancsmith:
I noticed we still have lots of links to our private repository. These links won't work for people outside of the sourcegraph organisation, so we update them to point to the public snapshot repository.

Note: there are other types of links we should probably update, especially around mentioned issues. Those are not tackled in this commit.


MaedahBatool (approved):
LGTM thanks a ton.

PR #803, opened 2024-11-15
Specify the need for classic PAT for indexing outside org's private repos

noah-berman:
Follow up to (this ticket)[https://sourcegraph.zendesk.com/agent/tickets/20672]. We don't specify that the PAT needs to be Classic, so I added a <callout> to make it more clear.


airamare01 (approved):
LGTM!

PR #849, opened 2024-12-12
chore(release): pg 16 update to docs

DaedalusG:
This PR updates some of our surface information about PG upgrades as well as cleans up some eager suggestions of the automatic upgrade upgrade path. 

It removes references of support for PG12 instead suggesting that we currently only support 16. This was intentional to trim the docs rather than threading out a lot of specific information about Sourcegraph versions.

### DIFF
--- a/docs/admin/postgres.mdx
+++ b/docs/admin/postgres.mdx
@@ -4,12 +4,11 @@ Sourcegraph uses several PostgreSQL databases to support various functionality. 
 
 - `pgsql` or `primary`: responsible for user data and account information
 - `codeintel-db`: provides support for lsif data and part of the code-intelligence
+- `codeinsights-db`: provides support for code insights data
 
 ## Version requirements
 
-We support any version **starting from 12**.
-
-> NOTE: ⚠️ Version **3.26** required only Postgres 9.6. You should check the [upgrade docs for your deployment type](/admin/updates/) as well as the [database upgrade docs](#upgrading-postgresql) below prior to upgrading to 3.27.
+We support postgres 16 and above. Older versions of Sourcegraph supported 12 see our [postgres 12 deprecation notice](/admin/postgres12_end_of_life_notice) for more details.
 
 ## Role requirements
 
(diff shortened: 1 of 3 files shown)


docs/admin/postgres.mdx
```diff
@@ -23,19 +22,23 @@ intarray
 pg_stat_statements
 pg_trgm
 pgcrypto
+vector
+pg_vector
```
keegancsmith, line 26: When I list in my devenv what extensions I have installed it only mentions `vector`. Is `pg_vector` the old name and you include it here just in case? Should we call out this distinction just below?

docs/admin/postgres.mdx
```diff
@@ -23,19 +22,23 @@ intarray
...
 ```
 
 ## [Configuring PostgreSQL](/admin/config/postgres-conf)
 
-## Upgrading PostgreSQL
+### Upgrading Built-in PostgreSQL
 
-Sourcegraph uses PostgreSQL as its main database and this documentation describes how to upgrade PostgreSQL between major versions.
+Sourcegraph deployments come packaged with containers using the `postgresql-16` image for `pgsql` and `codeintel-db` deployments and `postgresql-16-codeinsights` for `codeinsights`. These images contain an entryscript which will detect and upgrade postgres instances below version 12. See the (postgres 12 end of life notice)[/admin/prostgres12_end_of_life_notice#] for more details.
```
keegancsmith, line 33: minor: I believe it would only support upgrading from 12, not anything below 12?

docs/admin/postgres.mdx
jdpleiness: It should just be `vector` as that is what pg_vector shows up as.

jdpleiness, line 26:
```suggestion

```

docs/admin/postgres.mdx
```diff
@@ -23,19 +22,23 @@ intarray
...
 
-## Upgrading PostgreSQL
+### Upgrading Built-in PostgreSQL
 
-Sourcegraph uses PostgreSQL as its main database and this documentation describes how to upgrade PostgreSQL between major versions.
+Sourcegraph deployments come packaged with containers using the `postgresql-16` image for `pgsql` and `codeintel-db` deployments and `postgresql-16-codeinsights` for `codeinsights`. These images contain an entryscript which will detect and upgrade postgres instances below version 12. See the (postgres 12 end of life notice)[/admin/prostgres12_end_of_life_notice#] for more details. 
 
 > WARNING: Upgrading the PostgreSQL database requires stopping your Sourcegraph deployment which will result in **downtime**.
+>
+> Additionally, once started the upgrade process stopping the database container will corrupt sourcegraph data. We advise taking a backup before the upgrade.
```
jdpleiness, line 37:
```suggestion
> Additionally, once the upgrade process is started via the database container, interrupting the container before the upgrade is complete could result in corrupting the underlying Postgres database. We strongly advise taking a backup before the upgrade. 
```

docs/admin/postgres.mdx
```diff
@@ -23,19 +22,23 @@ intarray
...
 
 > WARNING: Upgrading the PostgreSQL database requires stopping your Sourcegraph deployment which will result in **downtime**.
+>
+> Additionally, once started the upgrade process stopping the database container will corrupt sourcegraph data. We advise taking a backup before the upgrade. 
 
+The `PG_UPGRADE_EXTRA_ARGS` env var can be set in the built-in `postgresql-16` and `postgresql-16-codeinsights` deployments to provide extra arguments to the `pg_upgrade` command. 
 
-
-> NOTE: If you running PostgreSQL externally, see [Upgrading external PostgreSQL instances](/admin/postgres#upgrading-external-postgresql-instances)
+Setting `PG_UPGRADE_EXTRA_ARGS=--link` provides the `--link` option to `pg_update`. Creating a hard link between the old and new data directories. This option is faster and requires less disk space, but if containers are killed during the upgrade, corrupted data will be unrecoverable without a backup.
```
jdpleiness, line 41:
```suggestion
Setting the env var `PG_UPGRADE_EXTRA_ARGS=--link` provides the `--link` option to the underlying `pg_update`; creating a hard links between the old and new data directories instead of copying data. This option is faster and requires less disk space, however if the database containers are interrupted during the upgrade process there is a strong possibility of data corruption. This can still be a viable choice for users who are confident in their container scheduling and backups and would like to avoid longer downtimes.

```

docs/admin/postgres.mdx
jdpleiness: > minor: I believe it would only support upgrading from 12, not anything below 12?

Yeah, that's correct. I don't _think_ we have any customers running anything older than 12 but we should be clear 😄

Chickensoupwithrice (approved):

jdpleiness:
Nit, but since it's docs and customer facing we should make everything `Postgres` as we have a few different spellings and capitalizations now.

DaedalusG:
@jdpleiness https://github.com/sourcegraph/docs/pull/852

PR #855, opened 2024-12-13
fix(rel): fix patch number, add Postgres notice

jdpleiness:
Fix patch number in changelog, add to latest release listing.


camdencheek (approved):

DaedalusG (approved):
Thanks for catching this! It should be fixed on release registry but I think maybe a new deploy needs to happen

DaedalusG:
https://github.com/sourcegraph/releaseregistry/pull/55

PR #871, opened 2024-12-20
Add a warning that auto-migrate to pg16 is not supported.

Chickensoupwithrice:
Relates to [REL-491](https://linear.app/sourcegraph/issue/REL-491/postgres-sourcegraph-server-aka-single-image-handles-postgres-16)
I think we should not support auto-migration of the database to pg16, since single container is not production capable deployment type, and the customers using it are explicitly trials.


MaedahBatool (approved):
TY

PR #896, opened 2025-01-15
feat: documentation for upgrading the single server container image to postgresql16

Chickensoupwithrice:
Completes documentation for a [PR](https://github.com/sourcegraph/sourcegraph/pull/2711)

## Pull Request approval

### DIFF
--- a/docs/admin/deploy/docker-single-container/index.mdx
+++ b/docs/admin/deploy/docker-single-container/index.mdx
@@ -138,6 +138,37 @@ SELECT * FROM users;
 
 ## Upgrade
 
+### Postgresql 16
+
+From sourcegraph version 5.11 onwards, the Sourcegraph single container Docker image uses Postgresql 16. Upgrading from Postgresql 12 to Postgresql 16 is a manual process, that is similar to the one outlined below for multi-version upgrades, but migrator has been merged into the container, allowing for a simpler upgrade.
+
+> NOTE: It is highly recommended to **take an up-to-date snapshot of your databases** prior to starting a multi-version upgrade. The upgrade process aggressively mutates the shape and contents of your database, and undiscovered errors in the migration process or unexpected environmental differences may cause an unusable instance or data loss.
+>
+> We recommend performing the entire upgrade procedure on an idle clone of the production instance and switch traffic over on success, if possible. This may be low-effort for installations with a canary environment or a blue/green deployment strategy.
+>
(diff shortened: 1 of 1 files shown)


docs/admin/deploy/docker-single-container/index.mdx
```diff
@@ -138,6 +138,37 @@ SELECT * FROM users;
...
+- Read our [update policy](/admin/updates/#update-policy) to learn about Sourcegraph updates.
+- Find the entries that apply to the version range you're passing through in the [update notes for Sourcegraph with Docker Single Container](/admin/updates/server#multi-version-upgrade-procedure).
+
+0. You must first shutdown the container instance via `docker stop [CONTAINER]`.
+1. Start a temporary Postgres container on top of the Postgres data directory used by the old `sourcegraph/server` image. You must use the *new* postgresql-16-codeinsights image, which is based on the new Postgresql 16 image, and provides an automatic upgrade script to move from Postgresql 12 to Postgresql 16.
+
+`docker run --rm -it -v ~/.sourcegraph/data/postgresql:/data/pgdata-12 -e POSTGRES_USER=postgres  -p 5432:5432 sourcegraph/postgresql-16-codeinsights:{CURRENT_VERSION_NO_V}`
+
+2. Once that temporary container marks that Postgresql has started, the migration is complete and you can stop the temporary container.
+3. You need to copy the `reindex.completed` file for the older reindex.
```
jdpleiness, line 162: The PG16 container _should_ create the reindex.completed file?

Chickensoupwithrice: It creates one with `5.10...` but the single server container checks against `5.1...`

jdpleiness: Ah got it. The reindex file is labeled 5.10 just cause that's when we switch to wolfi, but the version doesn't get read in or anything like that. In fact, until I changed it this time around, it was `3.5-reindex.completed` or something like that.

So maybe we can still use it, but you have more context so your call 👍

Chickensoupwithrice: I would say we should create both, but I'm not really sure where it's used or why the 5.10 one is created

Chickensoupwithrice:
Okay, with this new [PR](https://github.com/sourcegraph/sourcegraph/pull/2881), server now checks for 5.10 instead of 5.1, and the docs should reflect that

PR #898, opened 2025-01-15
Update and rename adding_lsif_to_workflows.mdx to adding_scip_to_work…

mibali:
…flows.mdx

Updated the dir from lsif to scip and updatd the github action workflow example link to point to the updated LSIF to SCIP link


airamare01 (approved):
GTG

PR #923, opened 2025-01-24
Preserve the path when redirecting the latest version

peterguy:
When redirecting links that contain the latest version in the path, only the slug was being preserved, and the redirect was going to sourcegraph.com.

Change that to preserve the whole path (except the version), and to preserve the current host.

For example:

`http://localhost:3000/docs/v/5.11/code_search/reference/queries` should redirect to `/docs/code_search/reference/queries`, which should finally redirect to `/docs/code-search/queries`

Behavior of `/@5.11/` should be the same as `/v/5.11/`.


peterguy:
Fixed elsewhere

PR #954, opened 2025-02-04
chore(changelog): remove leftover OPTIONAL; placeholders from changelog

burmudar:
Some changelog entries have `OPTIONAL; info at ...` which should actually be removed by an engineer when it gets merged but it is left over.


camdencheek (approved):

keegancsmith (commented):
we should get the changelog bot to complain if they leave this in (and if they devs don't make there beds in the morning).

PR #985, opened 2025-02-17
Configuring UID & GUID for native k8s executors

loujar:
added note to configure `KUBERNETES_RUN_AS_USER` & `KUBERNETES_RUN_AS_GROUP` for native k8s executors as it is now required to explicitly configure these ENVs when deploying native k8s executors

Follow up doc change for https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/618
## Pull Request approval


loujar:
Actually realizing this is not explicitly required since https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/618 set the default value to "blank", which removes the requirement to explicitly configure UID / GUID

> Having those settings commented out causes KUBERNETES_RUN_AS_USER and KUBERNETES_RUN_AS_GROUP in charts/sourcegraph-executor/k8s/templates/executor.ConfigMap.yaml to be blank, which is the new default value.

PR #987, opened 2025-02-18
Add callout for instances where NO_PROXY requires an asterisk

airamare01:
This change adds a nuanced item where some customers may need to add an asterisk when using proxies in their override file.

## Pull Request approval


jdpleiness (approved):
Thanks!

PR #992, opened 2025-02-20
De-purplize code blocks

toolmantim:
Before:

Screenshot 2025-02-20 at 3 59 02 pm

After:

Screenshot 2025-02-20 at 4 01 13 pm

Also cleaned up some of the type and positioning of the prose header block while I was there too.

### DIFF
--- a/src/components/CopyButton.tsx
+++ b/src/components/CopyButton.tsx
@@ -22,20 +22,20 @@ export function CopyButton({ text, lang }: { text: string, lang: string }) {
   }
 
   return (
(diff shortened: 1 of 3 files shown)


MaedahBatool:
hey @toolmantim thanks for contributing. Back then, we had some readability issues with code snippets, so we decided to use this code theme. Do you think the current theme is interfering with our brand guidelines?

toolmantim:
I think this is a small improvement for readability itself, but haven’t looked at what theme might be more on-brand.

PR #1021, opened 2025-03-10
feat: update docs for commit signing

bahrmichael:
Closes SRCH-1746

### DIFF
--- a/docs/admin/config/batch_changes.mdx
+++ b/docs/admin/config/batch_changes.mdx
@@ -30,11 +30,11 @@ To restore the default behavior, you can either delete the `batchChanges.rollout
 
 Or, to put it another way:
 
-| `batchChanges.rolloutWindows` configuration | Behavior |
-|---------------------------------------------|-----------|
-| Omitted, or set to `null`                   | Changesets will be reconciled as fast as the code host allows; essentially the same as setting a single `{"rate": "unlimited"}` window. |
+| `batchChanges.rolloutWindows` configuration |                                                                                                                                           Behavior                                                                                                                                           |
+| ------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Omitted, or set to `null`                   | Changesets will be reconciled as fast as the code host allows; essentially the same as setting a single `{"rate": "unlimited"}` window.                                                                                                                                                      |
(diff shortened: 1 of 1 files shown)


docs/admin/config/batch_changes.mdx
```diff
@@ -180,7 +180,31 @@ GitHub | API call | ✓ | ✓ |
...
 Gerrit | API call | ✗ | ✓ | Requires ["delete own changes" permission](https://gerrit-review.googlesource.com/Documentation/access-control.html#category_delete_own_changes) at minimum
 
-## Commit signing for GitHub
+## Commit signing
+
+Batch Changes supports Commit Signing with GitHub Apps since version 5.1 and with SSH Keys since version 6.2.
+
+### Commit signing with SSH Keys (GitLab and GitHub)
+
+<Callout type="note"> This feature is supported only in Sourcegraph versions 6.2 or more.</Callout>
```
derek-feriancek-sg, line 189: Nit: How about "or later" instead of "or more"?

derek-feriancek-sg: Hm, after finishing reading this diff I see we have "or more" in other parts of the docs as well. I'm curious what Maedah's stance is

bahrmichael: @MaedahBatool do we have a standard here?

bahrmichael:
@MaedahBatool This PR is now ready for review. The feature will be available to customers with the 6.2 release. Should we target main, or a different branch?

docs/admin/config/batch_changes.mdx
MaedahBatool: hey @bahrmichael we have the following formats:
- v6.2 or more
- v6.2+

docs/admin/config/batch_changes.mdx
```diff
@@ -180,7 +180,42 @@ GitHub | API call | ✓ | ✓ |
...
+
+Commit signing with SSH keys can only be used in combination with a personal access token.
```
LeweyM, line 191: Although we say here that only allow SSH keys for commit signing with PAT tokens, we allow it for fine-grained access tokens in the UI:

image

Is this something we need to correct in the UI, are the docs wrong, or am I getting this wrong?

MaedahBatool (approved):
Added few changes via direct commits. Approving now looks good :)

PR #1024, opened 2025-03-11
Add note about version to PCW docs

dadlerj:
This same version number applies to both VSC and JB right?

## Pull Request approval


kelsey-brown:
Yeah this is a good question - technically the final fixes for this logging that went out with that VSCode version happened in the _agent_, and were just inherited by JetBrains ([thread](https://linear.app/sourcegraph/issue/CODY-4196/metrics-ensure-characterslogger-is-accurate-in-jetbrains#comment-d6da5dc0) where I confirmed this previously). So there wasn't really a JB version associated officially...this was back before the releases were tied to one another directly

Looking at the [timing](https://github.com/sourcegraph/jetbrains/releases?page=1) the next JetBrains release that went out after VSCode 1.42 was 7.1.0 so we could add that filter here - but let me just double check with someone that that's how the version releases would work for changes made it the agent

kelsey-brown:
@dadlerj confirmed it's 7.1.0. FWIW, we have very few users running these versions so the filtering doesn't really make much of a difference to the PCW calc in aggregate across 30, 90, or 120 day timeframes (checked across 3 big customers). It's probably still worth updating for completeness though. 

Because of the way version numbers differ between clients, we'll need to implement an OR filter so we can handle the version filters for each client individually. Here's an example in Looker: https://sourcegraph.looker.com/explore/sourcegraph/v2_events?qid=fwTxOsbqxUl8lwz7bo10Kz&origin_space=562&toggle=dat,fil,vis

PR #1063, opened 2025-04-02
fix fragment links for `batch-spec-yaml-reference`

peterguy:
Links to fragments (bookmarks) in `batch-spec-yaml-reference` were incorrect: they contained dashes where dots are in the headings, but the bookmark generator (`GithubSlugger`) creates ids by removing alpha-numeric characters and encoding spaces as dashes, so the links with dashes where there were dots didn't work.

Remove all dashes; spot test the links.

Also did some general fixup of  `batch-spec-yaml-reference` links, where the dashes were underscores, and fixed a redirector that was directing to the wrong location.

Tacked on a similar fix in `docs/admin/code_hosts/bitbucket_server.mdx`


MaedahBatool (approved):
Fixed the merge conflicts. Let's make it live.

PR #1087, opened 2025-04-15
Update docs for Entra ID  to reference Azure Active Directory

noah-berman:
Make it really clear that Azure AD is the same thing as Entra


MaedahBatool (approved):
LGTM TY

PR #1092, opened 2025-04-22
Search Jobs: document new ENVs

stefanhengl:
Adds a table to document the ENVs we added in
https://github.com/sourcegraph/sourcegraph/pull/4975

Test plan:
N/A


janhartman (approved):
Do we also need to state that this is only true for 6.3+?

MaedahBatool (approved):
LGTM TY :)

stefanhengl:
> Do we also need to state that this is only true for 6.3+?

@MaedahBatool I think Jan's comment is still relevant, no? Do we have to add version information?

PR #1110, opened 2025-04-30
Add upgrade info for 6.4 repo updater removal

eseliger:
The service is no longer needed. Corresponding helm chart update: https://github.com/sourcegraph/deploy-sourcegraph-helm/pull/671

Test plan: n/a


DaedalusG:
We probably need to also review some of the helper commands etc in our run docs and make sure we're not having people run commands that reference `repo-updater` there as well. I'll have a look 👀

DaedalusG (approved):

PR #1117, opened 2025-05-06
Add 6.4 upgrade instruction for symbols

eseliger:
The symbols service will no longer be needed in 6.4. We still publish it at the moment, but it won't do anything. All functionality is now handled by searcher and frontend. Existing set env vars should be moved to the searcher service, and resources may be reallocated to there as well.

Test plan: We'll test this on dotcom and S2 for the next few weeks.


MaedahBatool (approved):
LGTM, @eseliger. But it's a draft. Is it still a WIP, or can I merge?

eseliger:
Moved it to review ready now! Wanted to talk to the team first :)

jdpleiness (approved):

PR #1121, opened 2025-05-06
Sourcegraph AMIs: remove references to AMI T-shirt sizes

DaedalusG:
We no longer publish images per instance T-shirt size. This PR removes those references from the deployment docs and moves them to a "how-to" so that they can still be referenced as needed.

Tested deployment via EC2 account


cloudyadi (approved):
thanks for quick turnaround on this

loujar (approved):

PR #1132, opened 2025-05-13
Cleanup documentation around symbols and repo-updater

eseliger:
Those two services are removed from our architecture, and workloads are moved to worker and searcher. This PR removes (hopefully) all remaining references to those services in our docs.

Test plan: Review.


unknwon (approved):
God bless you

PR #1148, opened 2025-05-21
modelconfig: document new OpenAI-compatible configurations

emidoots:
Documents the new `openaicompatible` model configurations supported as part of https://linear.app/sourcegraph/issue/CORE-1019/enablement-ga-self-hosted-models-additional-openai-compatible


MaedahBatool (approved):
Hi @emidoots, I fixed a few build and formatting errors and added commits to your branch. I feel we are good to go. Let's ship it, approved! :)

PR #1211, opened 2025-06-25
GitHub Enterprise Apps docs

pjlast:
Adds some docs on how to use GitHub Enterprise Apps (i.e. add the GitHub App manually)

Also mention how to set up an app-wide code host connection

### DIFF
--- a/docs/admin/code_hosts/github.mdx
+++ b/docs/admin/code_hosts/github.mdx
@@ -140,7 +140,6 @@ To add an existing GitHub App:
    - **GitHub URL**: The URL of your GitHub instance (e.g., `https://github.com` or `https://github-enterprise.example.com`)
    - **Client ID**: The unique identifier for your GitHub App
    - **Private Key**: The private key generated for your GitHub App (in PEM format)
-
 4. Click **Add GitHub App** to save the configuration.
 5. Once added, you can manage installations and create code host connections just like with apps created through Sourcegraph.
 
--- a/docs/admin/code_hosts/github.mdx
+++ b/docs/admin/code_hosts/github.mdx
@@ -22,8 +22,8 @@ There are 2 ways to connect with GitHub:
 {/* <!-- NOTE: The instructions in the following sections closely mirror those in doc/admin/config/batch_changes under "Commit signing for GitHub". When making changes here, be sure to consider if those changes should also be made over there! --> */}
 
 There are two ways to connect a GitHub App:
-1. **Create a new GitHub App** via Sourcegraph, which will set up all of the required permissions and automatically retrieve the credentials
-2. **Add an existing GitHub App** by providing the details of an existing GitHub App (required when using GitHub Enterprise Apps)
+1. **[Create a new GitHub App](#creating-a-new-github-app)** via Sourcegraph, which will set up all of the required permissions and automatically retrieve the credentials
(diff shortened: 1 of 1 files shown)


docs/admin/code_hosts/github.mdx
```diff
@@ -452,6 +494,8 @@ GitHub connections support the following configuration options, which are specif
 	//   ]
 
 	// If non-null, this is a GitHub App connection with some additional properties.
+	// To sync repositories from all installations of a GitHub App, omit the installationID field.
+	// To sync repositories from a specific installation only, include the installationID field.
 	"gitHubAppDetails": null,
```
eseliger: this is the most useless piece of documentation btw 😆 the JSON schema renderer is so bad lol

docs/admin/code_hosts/github.mdx
```diff
@@ -21,6 +21,12 @@ There are 2 ways to connect with GitHub:
 
 {/* <!-- NOTE: The instructions in the following sections closely mirror those in doc/admin/config/batch_changes under "Commit signing for GitHub". When making changes here, be sure to consider if those changes should also be made over there! --> */}
 
+There are two ways to connect a GitHub App:
+1. **[Create a new GitHub App](#creating-a-new-github-app)** via Sourcegraph, which will set up all of the required permissions and automatically retrieve the credentials
+2. **[Add an existing GitHub App](#adding-an-existing-github-app)** by providing the details of an existing GitHub App (required when using GitHub Enterprise Apps)
```
eseliger, line 26: just wondering, is it theoretically possible to create a GHE App via the post method as well, we just don't support it right now? 

Maybe we could look into that one day.

docs/admin/code_hosts/github.mdx
```diff
@@ -111,10 +125,38 @@ Once public, App can be installed in additional namespaces either from Sourcegra
...
+To add an existing GitHub App:
+
+1. Go to **Site admin > Repositories > Github Apps** on Sourcegraph.
+2. Click **Add an existing GitHub App**.
+3. Enter the following details of your existing GitHub App:
+   - **GitHub URL**: The URL of your GitHub instance (e.g., `https://github.com` or `https://github-enterprise.example.com`)
+   - **Client ID**: The unique identifier for your GitHub App
+   - **Private Key**: The private key generated for your GitHub App (in PEM format)
+4. Click **Add GitHub App** to save the configuration.
+5. Once added, you can manage installations and create code host connections just like with apps created through Sourcegraph.
```
eseliger, line 144: do we need to call out here what we provision for GitHub Apps usually, so the webhook events, token scopes, etc are sufficient? Plus creating a client id/secret pair and adding it to auth.providers

docs/admin/code_hosts/github.mdx
pjlast: I know 😭 but it's where it would be :(

pjlast: But it looks bad idk I'll remove it

docs/admin/code_hosts/github.mdx
pjlast: Nothing in the docs makes me believe that it's not possible, but the one customer that tried it (that we know of, at least) could not get it to work

eseliger: we probably don't post the form to the right endpoint or so, the enterprise GH Apps page is in a completely different place 😬

pjlast: I'll have a look! Maybe we can fix that

PR #1237, opened 2025-07-02
analytics: fix information hierarchy

bobheadxi:
- Analytics is not a Cloud feature, no reason to name it and the pages "Cloud"
- Move API to another page - FYI @jac 
- Remove repetitive information


unknwon (approved):
Thank you!

PR #1258, opened 2025-07-11
No byok for Deep Search

mmanela:
Update Deep Search docs to explain we don't support BYOK


jamesmcnamara (approved):

sanjayesn (approved):

MaedahBatool:
> @MaedahBatool Should the notice of no BYOK support be in a callout?

Yes we can add one callout.

PR #1304, opened 2025-08-01
update format of HTTP PROXY ENV config

loujar:
Updated config format for HTTP PROXY ENV examples in this doc for compatibility with (and avoid overwritting) default helm value ENVs


marcleblanc2 (approved):
LGTM

PR #1326, opened 2025-08-14
chore: Some cleanup and restructuring

eseliger:
Split self-hosted stuff from stuff all admins are concerned about. Kill some dead pages, remove internal docs.

Test plan: Review.


unknwon (approved):

bobheadxi (approved):
ship it!

PR #1327, opened 2025-08-14
oauth: Add documentation on OAuth integrations and M2M credentials

eseliger:
Closes SRC-1356.

Test plan: Review.

### DIFF
--- a/docs/admin/oauth_apps.mdx
+++ b/docs/admin/oauth_apps.mdx
@@ -327,7 +327,7 @@ curl -X POST https://sourcegraph.example.com/.auth/idp/oauth/token \
   -d "client_secret=YOUR_CLIENT_SECRET" # Only for private clients
 ```
 
-**Important**: Always store the new refresh token returned in the response, as it may differ from the one you used in the request.
+**Important**: Always store the new refresh token returned in the response, as it may differ from the one you used in the request. Every refresh token is strictly one-time use, your application needs to handle the case when it's possible to have multiple processes initiating the token refresh process because only one of them will succeed for the same refresh token.
 
 ### Token Expiration Handling
 


public/llms.txt
bobheadxi, line 1: is this diff intentional?

bobheadxi (approved):

public/llms.txt
eseliger: yeah I regenerated this file

pjlast (approved):

docs/admin/oauth_apps.mdx
```diff
@@ -0,0 +1,352 @@
+# OAuth Apps
+
+OAuth Apps let you build rich third-party experiences that call Sourcegraph **on behalf of end-users while honoring their repository permissions**. Every API request made with an OAuth access token is evaluated against the same authorization rules that apply inside the Sourcegraph UI, so your integration only sees the code each user is allowed to see.
```
unknwon: nit

```suggestion
OAuth Apps let you build rich third-party experiences that interact with Sourcegraph **on behalf of end-users while honoring their repository permissions**. Every API request made with an OAuth access token is evaluated against the same authorization rules that apply inside the Sourcegraph UI, so your integration only sees the code each user is allowed to see.
```

docs/admin/oauth_apps.mdx
```diff
@@ -0,0 +1,352 @@
...
+
+This makes OAuth the preferred choice for any multi-user integration like browser extensions, IDE plugins, chatbots, internal tools, or SaaS platforms where you need to:
+
+- Authenticate users with familiar Sourcegraph credentials
+- Operate without ever handling or storing users' passwords
+- Rely on short-lived, scoped tokens that can be revoked at any time
+- Guarantee that repository visibility and other access controls are enforced automatically
+
+OAuth Apps follow standards [RFC 6749](https://tools.ietf.org/html/rfc6749), [RFC 7636 (PKCE)](https://tools.ietf.org/html/rfc7636), and [RFC 8628 (Device Authorization Grant)](https://tools.ietf.org/html/rfc8628).
```
unknwon: nit

```suggestion
OAuth Apps are in compliance with standards including [RFC 6749](https://tools.ietf.org/html/rfc6749), [RFC 7636 (PKCE)](https://tools.ietf.org/html/rfc7636), and [RFC 8628 (Device Authorization Grant)](https://tools.ietf.org/html/rfc8628).
```

docs/admin/oauth_apps.mdx
```diff
@@ -0,0 +1,352 @@
...
+| Device Authorization | `https://sourcegraph.example.com/.auth/idp/oauth/device/code` |
+| Token Revocation | `https://sourcegraph.example.com/.auth/idp/oauth/revoke` |
+| Token Introspection | `https://sourcegraph.example.com/.auth/idp/oauth/introspect` |
+| User Info | `https://sourcegraph.example.com/.auth/idp/oauth/userinfo` |
+
+## OAuth Flow Examples
+
+<Accordion title="Authorization Code Flow (Private Clients)">
+
+1. **Authorization Request**: Redirect users to Sourcegraph's authorization endpoint:
```
unknwon, line 98: nit

```suggestion
1. **Authorization Request**: Redirect users to Sourcegraph's authorization endpoint (line breaks for better illustration):
```

docs/admin/oauth_apps.mdx
```diff
@@ -0,0 +1,352 @@
...
+  response_type=code&
+  client_id=YOUR_CLIENT_ID&
+  redirect_uri=YOUR_REDIRECT_URI&
+  scope=user:all&
+  state=RANDOM_STATE_STRING
+```
+
+The user may be asked for consent to grant your app the requested scopes if not previously granted.
+
+2. **Authorization Response**: Sourcegraph redirects back with an authorization code and the state parameter:
```
unknwon, line 111: nit

```suggestion
2. **Authorization Response**: Sourcegraph redirects back with an authorization code and the state parameter (line breaks for better illustration):
```

docs/admin/oauth_apps.mdx
```diff
@@ -0,0 +1,352 @@
...
+```bash
+curl -X POST https://sourcegraph.example.com/.auth/idp/oauth/token \
+  -H "Content-Type: application/x-www-form-urlencoded" \
+  -d "grant_type=refresh_token" \
+  -d "refresh_token=YOUR_REFRESH_TOKEN" \
+  -d "client_id=YOUR_CLIENT_ID" \
+  -d "client_secret=YOUR_CLIENT_SECRET" # Only for private clients
+```
+
+**Important**: Always store the new refresh token returned in the response, as it may differ from the one you used in the request.
```
unknwon:
```suggestion
**Important**: Always store the new refresh token returned in the response, as it may differ from the one you used in the request. Every refresh token is strictly one-time use, your application needs to handle the case when it's possible to have multiple processes initiating the token refresh process because only one of them will succeed for the same refresh token.
```

unknwon (approved):
Thank you!

docs/admin/oauth_apps.mdx
eseliger: Do you think someone who we want to trust to implement OAuth correctly will not understand that? 😬

PR #1351, opened 2025-09-08
docs(sourcegraph): add docs for the MCP server

jamesmcnamara:
Adds documentation for the Sourcegraph MCP server added in v6.8

## Pull Request approval

### DIFF
--- a/docs/api/mcp/index.mdx
+++ b/docs/api/mcp/index.mdx
@@ -1,0 +1,246 @@
+# Sourcegraph MCP Server
+
+<p className="subtitle">Connect AI agents and applications to your Sourcegraph instance's code search and analysis capabilities.</p>
+
+<TierCallout>
+  Supported on [Enterprise](/pricing/enterprise) plans.
+</TierCallout>
+<Callout type="note">This feature is [experimental](admin/beta_and_experimental_features#experimental-features) and might change or be removed in the future.</Callout>
+
+The Sourcegraph Model Context Protocol (MCP) Server provides AI agents and applications with programmatic access to your Sourcegraph instance's code search, navigation, and analysis capabilities through a standardized interface.
+
+## Server Endpoint
+
+The MCP server is available at:
+
+```
+https://your-sourcegraph-instance.com/.api/mcp/v1
+```
+
+## Authentication
+
+The Sourcegraph MCP server supports two authentication methods:
+
+### OAuth 2.0
+
+For programmatic access without storing credentials, use OAuth 2.0 with device flow or authorization code flow with PKCE:
+
+#### Prerequisites
+
+Before using OAuth, you must create an OAuth application in your Sourcegraph instance. Follow the instructions [here](/admin/oauth_apps#creating-an-oauth-app). (Note: you will need the `user:all` scope) 
+
+### Authorization Header
+
+Include your token in the Authorization header:
+
+```
+Authorization: token YOUR_ACCESS_TOKEN
+```
+
+## Available Tools
+
+The MCP server provides these tools for code exploration and analysis:
+
+### File & Repository Operations
+
+#### `sg_read_file`
+
+Read file contents with line numbers and support for specific ranges and revisions.
+
+**Parameters:**
+
+- `repo` - Repository name (required)
+- `path` - File path within repository (required)
+- `startLine` - Starting line number (optional)
+- `endLine` - Ending line number (optional)
+- `revision` - Branch, tag, or commit hash (optional)
+
+**Use cases:** Reading specific files, examining code sections, reviewing different versions
+
+<Callout type="note">File size limit is 128KB. Use line ranges for larger files.</Callout>
+
+#### `sg_list_files`
+
+List files and directories in a repository path.
+
+**Parameters:**
+- `repo` - Repository name (required)
+- `path` - Directory path (optional, defaults to root)
+- `revision` - Branch, tag, or commit hash (optional)
+
+#### `sg_list_repos`
+
+Search and list repositories by name patterns with pagination support.
+
+**Parameters:**
+- `query` - Search pattern for repository names (required)
+- `limit` - Maximum results per page (optional, default 50)
+- `after`/`before` - Pagination cursors (optional)
+
+### Code Search
+
+#### `sg_keyword_search`
+
+Perform exact keyword searches with boolean operators and filters.
+
+**Parameters:**
+- `query` - Search query with optional filters (required)
+
+**Supported filters:**
+- `repo:` - limit to specific repositories
+- `file:` - search specific file patterns
+- `rev:` - search specific revisions
+
+**Features:** Boolean AND/OR operators, regex patterns
+
+#### `sg_nls_search`
+
+Semantic search with flexible linguistic matching.
+
+**Parameters:**
+- `query` - Natural language search query (required)
+
+**Supported filters:**
+- `repo:` - limit to specific repositories
+- `file:` - search specific file patterns
+- `rev:` - search specific revisions
+
+**Features:** Flexible linguistic matching, stemming, broader results than keyword search
+
+### Code Navigation
+
+#### `sg_go_to_definition`
+
+Find the definition of a symbol from a usage location.
+
+**Parameters:**
+- `repo` - Repository name (required)
+- `path` - File path containing symbol usage (required)
+- `symbol` - Symbol name to find definition for (required)
+- `revision` - Branch, tag, or commit hash (optional)
+
+**Features:** Cross-repository support, compiler-level accuracy
+
+#### `sg_find_references`
+
+Find all references to a symbol from its definition location.
+
+**Parameters:**
+- `repo` - Repository name (required)
+- `path` - File path containing symbol definition (required)
+- `symbol` - Symbol name to find references for (required)
+- `revision` - Branch, tag, or commit hash (optional)
+
+### Version Control & History
+
+#### `sg_commit_search`
+
+Search commits by message, author, content, files, and date ranges.
+
+**Parameters:**
+- `repos` - Array of repository names (required)
+- `messageTerms` - Terms to search in commit messages (optional)
+- `authors` - Filter by commit authors (optional)
+- `contentTerms` - Search in actual code changes (optional)
+- `files` - Filter by file paths (optional)
+- `after`/`before` - Date range filters (optional)
+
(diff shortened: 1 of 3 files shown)


docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+Authorization: token YOUR_ACCESS_TOKEN
+```
+
+### OAuth 2.0 Device Flow
+
+For programmatic access without storing credentials, use the OAuth 2.0 device flow:
+
+#### Prerequisites
+
+Before using the device flow, you must create an OAuth application in your Sourcegraph instance:
```
mmanela: You can link to the official oauth docs that @eseliger added here

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+#### `sg_list_repos`
+
+Search and list repositories by name patterns with pagination support.
+
+**Parameters:**
+- `query` - Search pattern for repository names (required)
+- `limit` - Maximum results per page (optional, default 50)
+- `after`/`before` - Pagination cursors (optional)
+
+### Code Search
```
mmanela, line 80: What do you think @janhartman ?

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+#### `sg_get_contributor_repos`
+
+Find repositories where a contributor has made commits.
+
+**Parameters:**
+- `author` - Author name or email (required)
+- `limit` - Maximum repositories to return (optional, default 20)
+- `minCommits` - Minimum commits required (optional, default 1)
+
+### Deep Search
```
janhartman: Let's link to the DS page here maybe

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+**Supported filters:**
+- `repo:` - Limit to specific repositories
+- `file:` - Search specific file patterns
+- `rev:` - Search specific revisions
+
+**Features:** Boolean AND/OR operators, regex patterns
+
+#### `sg_nls_search`
+
+Natural language semantic search for conceptual code queries.
```
janhartman: This isn't really a completely semantic search 😅 The only semantic component is the reranker at the end, whereas the matching is still keyword-based (just with OR instead of AND between terms). Yeah, I know the name isn't that good 😄

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+**Parameters:**
+- `query` - Natural language search query (required)
```
janhartman, line 101: NLS search supports the same filters as above (repo, file, rev), let's add them here

docs/api/mcp/index.mdx
janhartman: Let's change this line to 'Semantic search with flexible lingusitc matching'.

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
+# Sourcegraph MCP Server
+
+<p className="subtitle">Connect AI agents and applications to your Sourcegraph instance's code search and analysis capabilities.</p>
+
+The Sourcegraph Model Context Protocol (MCP) Server provides AI agents and applications with programmatic access to your Sourcegraph instance's code search, navigation, and analysis capabilities through a standardized interface.
+
+<Callout type="info">MCP Server requires Sourcegraph 6.8+ and is available on all deployment types.</Callout>
```
eseliger: I don't think we need to call out either of those.

1. The docs are versioned and clearly say 6.8 in the top right corner
2. If ALL deployment types work, why mention it?

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+```
+https://your-sourcegraph-instance.com/.api/mcp/v1
+```
+
+## Authentication
+
+The Sourcegraph MCP server supports three authentication methods:
+
+### Access Token (Query Parameter)
```
eseliger: Should we recommend this? query parameters are generally considered not sensitive so those can even show up in Cloudflare logs and things 🙃 cc @evict if you have any thoughts on this.

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+### Authorization Header
+
+Include your token in the Authorization header:
+
+```
+Authorization: token YOUR_ACCESS_TOKEN
+```
+
+### OAuth 2.0 Device Flow
```
eseliger: Hmm, I'm not sure device flow is required here, that just depends on the MCP client -- auth code flow with PKCE can also be fine. So maybe just call it OAuth 2.0?

Also, let's make OAuth 2.0 the first option in the list, as it's the most spec-compliant and most secure option

docs/api/mcp/index.mdx
eseliger: We can [predefine oauth clients](https://sourcegraph.sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fsourcegraph&branch=main&file=cmd%2Ffrontend%2Finternal%2Fauth%2Fidp%2Fdatabase%2Fpredefined_clients.go&editor=VSCode&version=2.4.0&start_row=12&start_col=0&end_row=20&end_col=0) as well, I wonder if we should do that.

@unknwon do you have thoughts here?

Then users would just need to copy the details here. No way to get a client secret then, so PKCE or RFC8628 device flow auth are required; which is probably what most clients should do anyways.

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+For programmatic access without storing credentials, use the OAuth 2.0 device flow:
+
+#### Prerequisites
+
+Before using the device flow, you must create an OAuth application in your Sourcegraph instance:
+
+1. Navigate to `https://your-sourcegraph-instance.com/site-admin/oauth-clients/new`
+2. Complete the OAuth client creation wizard:
```
eseliger: we should talk about scopes here.

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+   POST /.auth/idp/oauth/device/code
+   ```
+
+2. **Poll for token:**
+   ```
+   POST /.auth/idp/oauth/token
+   ```
+
+3. **User authorization:**
+   Visit the provided URL to authorize the device.
```
eseliger: this is already outlined in oauth docs, no need to talk about this here, and generally this is information for MCP client implementors so probably can just omit this entirely?

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
...
+2. **Progressive Search:** Start with broad searches (`sg_nls_search`) then narrow with specific tools
+3. **File Verification:** Use `sg_list_files` before `sg_read_file` to verify file existence
+4. **Pagination:** Use `after`/`before` cursors for large result sets
+5. **Tool Combinations:** Chain tools together (e.g., `sg_list_repos` → `sg_commit_search`)
+
+## Related
+
+- [Sourcegraph GraphQL API](/api/graphql/)
+- [Sourcegraph Stream API](/api/stream_api/)
+- [Code Search](/code-search/)
```
eseliger: do we think this section helps someone? (I like to keep docs clean of extra content 😄)

docs/api/mcp/index.mdx
evict: Let's not recommend that! It shows up in plain Cloudflare logs. We should probably look into removing this auth method entirely.

docs/api/mcp/index.mdx
unknwon: My 2 cents: If we are aiming enterprises to use MCP feature to start with, let's ask users to create OAuth applications for now because our OAuth scopes are currently still all or nothing. Ideally, we could give away some less sensitive scopes OOTB for a predefined client, then if users want to grant more scopes, they can choose to create a different OAuth client for their MCP needs.

eseliger: Alright!

docs/api/mcp/index.mdx
jamesmcnamara: This is currently the only way to add an MCP server to Amp and I doubt we'll add another one before 6.8

jamesmcnamara:
> Seems pretty good! I wonder if we should generate full descriptions from our code and put them in the docs (collapsed by default). Could be valuable for an interested reader as otherwise they'd have to connect to the server first. But maybe hiding all the details is a feature? 😆

@janhartman What do you mean by full descriptions? Do you mean including the full tool description here (but collapsed)?

docs/api/mcp/index.mdx
eseliger: Still not great to recommend this IMO, we should implement custom headers or OAuth in Amp then, there's currently nothing that treats the URLs as particularly sensitive I believe so easy to mishandle the URL containing the token. 

This feels like a useful investment generally to support other tools as well. :)

docs/api/mcp/index.mdx
```diff
@@ -0,0 +1,273 @@
+# Sourcegraph MCP Server
```
eseliger, line 1: maybe for consistency we should add this

```
<TierCallout>
  Supported on [Enterprise](/pricing/enterprise) plans.
  <user>
    Currently available via the Web app.
  </user>
</TierCallout>
```

(other feature pages have it, too. not sure on the Web app part..)

do we allow it on Enterprise Starter? I guess yes?

eseliger: also somewhere close to the top of this page, we should add a notice about the product stage of the MCP server - since it's default on but I believe we settled on not GA yet it would be Beta: https://sourcegraph.com/docs/admin/beta_and_experimental_features#beta-and-experimental-features

mmanela: Lets call it as Experimental

jamesmcnamara: Done

docs/api/mcp/index.mdx
jamesmcnamara: Removed

docs/api/mcp/index.mdx
jamesmcnamara: Linked!

docs/api/mcp/index.mdx
jamesmcnamara: Removed

PR #1361, opened 2025-09-22
Add support for GPT-5 models in documentation

kalanchan:
add gpt-5 models to supported models page


emidoots:
Agreed, auto generation of this makes sense.

PR #1394, opened 2025-10-24
GitHub Apps: Add additional permissions requirement

pjlast:
Add an additional required permission to GitHub Apps, which is required when using groups caching


docs/admin/code_hosts/github.mdx
```diff
@@ -47,6 +47,7 @@ To create a GitHub App and connect it to Sourcegraph:
     Emails (Manage a user's email addresses): read
     Members (Organization members and teams): read
     Metadata (Search repositories, list collaborators, and access repository metadata): read
+    Administration (Repository creation, deletion, settings, teams, and collaborators): read (required for groups caching)
```
eseliger: is the idea to turn on the caching for everyone soon? Otherwise might be an easier sell for now to say


```suggestion
    Administration (Repository creation, deletion, settings, teams, and collaborators): read (only required for groups caching)
```

?

PR #1470, opened 2025-12-09
fix: Change page title to "Sourcegraph Cloud"

taiyab:


eseliger (approved):
hmm the purpose is not to introduce cloud here I think it's just the "Getting started with Sourcegraph" guide for the customers that are on cloud (so it omits some stuff that is on-prem specific)

but approving to unblock further changes

PR #1562, opened 2026-01-26
Add daily broken link checker with Slack reporting

eugenio-sanchez-sg:
## Problem

We have internal links in our docs that point to pages that have been removed or renamed. 

## Solution

A GitHub Action that runs daily at 6am UTC, scans all docs for broken internal links, and reports the results to a Slack channel. 

## Implementation

**New files:**
- `scripts/check-internal-links.js` — Node script that:
  - Builds a set of valid routes from all MDX files
  - Extracts internal links from markdown `[text](/path)` and JSX `href="/path"` patterns
  - Skips code blocks to avoid false positives
  - Detects case mismatches (links that work on macOS but fail on Linux)
  - Reports to Slack via webhook (or prints locally if no webhook)

- `.github/workflows/check-links.yml` — Cron workflow that runs daily and sends results to Slack

**Usage:**
```bash
npm run check-internal-links  # Run locally (prints to console)
```

## Setup required (after merge)

1. Create a Slack Incoming Webhook at `api.slack.com/apps`
2. Add `SLACK_WEBHOOK_URL` to GitHub repo **Settings → Secrets → Actions**

## Open questions

- Is daily the right cadence, or would weekly be better?
- Which Slack channel should receive these reports?

Feedback welcome — this is a proposal to see if this approach makes sense for the team.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -1,0 +1,24 @@
+name: Daily Link Check
+
+on:
+  schedule:
+    - cron: '0 6 * * *' # 6am UTC daily
+  workflow_dispatch: # Manual trigger
+
+jobs:
+  check-links:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install dependencies
+        run: npm install glob
+
+      - name: Check internal links
+        run: node scripts/check-internal-links.js
+        env:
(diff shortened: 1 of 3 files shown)


marcleblanc2:
I'd recommend we keep, improve, and reuse the existing https://github.com/sourcegraph/docs/blob/main/dev/check-links.mjs script, rather than duplicate it. It can be reused in a GitHub Actions cron job, to ping us in Slack with a list of broken links each day.

marcleblanc2:
Closing as part of the stale-branch cleanup (see #1950); the branch is being deleted. Reopen from a fresh branch if still needed.

PR #1598, opened 2026-02-12
Aws ftr updates

erclm:
Adding AWS Required sections for FTR


DaedalusG (changes requested):
Blocking this for now. AMI deployments are being sunset in the coming 7.0.0 release and beyond that we need to review the ARR that comes in from Amazon marketplace credits before we commit any further to supporting these deployment paths

DaedalusG (approved):
Approving after conversations at merge, note that ami/machine images will be sunset in 7.0.0

PR #1745, opened 2026-04-10
docs: add clarification between oauth apps and M2M credentials

enriquegh:


docs/admin/service-accounts.mdx
```diff
@@ -36,7 +44,11 @@ For detailed information about creating, managing, and using traditional access
 
 ### M2M OAuth Credentials (Client Credentials Flow)
 
-M2M credentials provide short-lived tokens via the OAuth client credentials flow expiring after 1 hour.
+M2M credentials provide short-lived tokens via the OAuth client credentials flow expiring after 1 hour. Tokens issued by this flow represent the service account itself, not an end-user session.
+
+If your instance's OpenID Connect discovery document lists `client_credentials`
+in `grant_types_supported`, that means the Sourcegraph OAuth server supports
+that grant globally. It does not mean a regular OAuth App can use it.
```
eseliger: hmm this confuses me, it sounds like some instances would advertise we don't support client credentials?

eseliger (approved):

docs/admin/service-accounts.mdx
enriquegh: yeah true. did a small change to clarify that this is for all instances

PR #1747, opened 2026-04-10
object-storage: add new 'sourcegraph' bucket

bobheadxi:
See https://github.com/sourcegraph/sourcegraph/pull/11613, https://github.com/sourcegraph/sourcegraph/pull/11614

Preview: https://sourcegraph-docs-git-new-sourcegrap-ffb85d-sourcegraph-f8c71130.vercel.app/self-hosted/external-services/object-storage

Closes https://linear.app/sourcegraph/issue/SVC-2259/sourcegraphobjects-on-prem-guidance


docs/self-hosted/external-services/object-storage.mdx
```diff
@@ -4,115 +4,194 @@ By default, Sourcegraph will use a `sourcegraph/blobstore` server bundled with t
...
+</Callout>
+
+<a id="sourcegraph-bucket" />
+
+## `sourcegraph` bucket
+
+<Callout type="warning">
+	Starting in [Sourcegraph 7.2](https://sourcegraph.com/changelog/releases/7.2), self-hosted Sourcegraph instances using S3 or
+	GCS object storage should now provision an additional bucket named
+	[`sourcegraph`](#sourcegraph-bucket). Sourcegraph currently reports a warning
```
eseliger: this to me sounds like it has to be called sourcegraph in GCS/S3 but the namespace of bucket names is global IIRC so only one project can ever claim this.

docs/self-hosted/external-services/object-storage.mdx
```diff
@@ -4,115 +4,194 @@ By default, Sourcegraph will use a `sourcegraph/blobstore` server bundled with t
...
+The `sourcegraph` bucket is intended to be the single bucket for new Sourcegraph features. Instead of creating one bucket per feature, new features store objects under namespaced key prefixes within this bucket.
+
+Existing buckets for code graph indexes and search jobs remain in use. This change ensures future features can be enabled without requiring a new bucket for each feature.
+
+### Using S3 for the `sourcegraph` bucket
+
+Set the following environment variables to target an S3 bucket for shared Sourcegraph uploads.
+
+- `SOURCEGRAPH_UPLOAD_BACKEND=S3`
+- `SOURCEGRAPH_UPLOAD_BUCKET=sourcegraph` (default)
```
eseliger: is this actually the default? I feel like this can't be defaulted as the name is globally unique

docs/self-hosted/external-services/object-storage.mdx
```diff
@@ -4,115 +4,194 @@ By default, Sourcegraph will use a `sourcegraph/blobstore` server bundled with t
...
+- `SOURCEGRAPH_UPLOAD_AWS_ENDPOINT=https://s3.us-east-1.amazonaws.com`
+- `SOURCEGRAPH_UPLOAD_AWS_ACCESS_KEY_ID=<your access key>`
+- `SOURCEGRAPH_UPLOAD_AWS_SECRET_ACCESS_KEY=<your secret key>`
+- `SOURCEGRAPH_UPLOAD_AWS_SESSION_TOKEN=<your session token>` (optional)
+- `SOURCEGRAPH_UPLOAD_AWS_USE_EC2_ROLE_CREDENTIALS=true` (optional; set to use EC2 metadata API over static credentials)
+- `SOURCEGRAPH_UPLOAD_AWS_USE_PATH_STYLE=false` (optional)
+- `SOURCEGRAPH_UPLOAD_AWS_REGION=us-east-1` (default)
```
eseliger: feels weird to default to some AWS region 🤔 if anything I'd expect that it does metadata server discovery to find the right region

docs/self-hosted/external-services/object-storage.mdx
```diff
@@ -4,115 +4,194 @@ By default, Sourcegraph will use a `sourcegraph/blobstore` server bundled with t
...
+
+### Using GCS for the `sourcegraph` bucket
+
+Set the following environment variables to target a GCS bucket for shared Sourcegraph uploads.
+
+- `SOURCEGRAPH_UPLOAD_BACKEND=GCS`
+- `SOURCEGRAPH_UPLOAD_BUCKET=sourcegraph` (default)
+- `SOURCEGRAPH_UPLOAD_GCP_PROJECT_ID=<my project id>`
+- `SOURCEGRAPH_UPLOAD_GOOGLE_APPLICATION_CREDENTIALS_FILE=</path/to/file>`
+- `SOURCEGRAPH_UPLOAD_GOOGLE_APPLICATION_CREDENTIALS_FILE_CONTENT=<{"my": "content"}>`
```
eseliger: can we use instance roles too? I feel like within GKE it should not be required to set this explicitly. 🤔

docs/self-hosted/external-services/object-storage.mdx
```diff
@@ -4,115 +4,194 @@ By default, Sourcegraph will use a `sourcegraph/blobstore` server bundled with t
...
+
+### Automatically provision the `sourcegraph` bucket
```
eseliger: should we even have that? I wonder if anyone really uses this.

docs/self-hosted/external-services/object-storage.mdx
bobheadxi: Hm, they all have defaults: https://sourcegraph.sourcegraph.com/r/github.com/sourcegraph/sourcegraph@main/-/blob/internal/codeintel/shared/lsifuploadstore/config.go?L36

docs/self-hosted/external-services/object-storage.mdx
bobheadxi: This is all copy-pasta from existing docs 😅 https://sourcegraph.com/docs/self-hosted/external-services/object-storage#using-s3-1

But yeah good callouts, I'll see what I can find

docs/self-hosted/external-services/object-storage.mdx
bobheadxi: Yeah, that's what we do in Cloud

will add as an option

docs/self-hosted/external-services/object-storage.mdx
bobheadxi: It's hard-required for `blobstore` (i.e. you can't turn it off) but yeah, not sure either. Currently it doesn't seem like significant complexity to retain

docs/self-hosted/external-services/object-storage.mdx
bobheadxi: Hmm, yeah - all the other backends have a default name 🤔 

I think the default is more for blobstore for anything else, but the existing docs all explicitly mention the default. I think for non-blobstore I'll just tweak the wording here

docs/self-hosted/external-services/object-storage.mdx
bobheadxi: Updated

docs/self-hosted/external-services/object-storage.mdx
bobheadxi: Updated

docs/self-hosted/external-services/object-storage.mdx
bobheadxi: I ended up adding this back, the default is meaningful and you need to set the endpoint yourself as well if you change it

eseliger (approved):

devdinu (approved):

filiphaftek (approved):

PR #1770, opened 2026-05-08
fix(batch-changes): repair broken permissions table

eugenio-sanchez-sg:
## What

Fixes the broken table under "Batch Change access for each permission level" on
`/batch-changes/permissions-in-batch-changes`. Previously the table rendered as a
single paragraph with literal `|` separators visible to the reader.

## Currently at Permissions in Batch Changes

You can see the page [here](https://sourcegraph.com/docs/batch-changes/permissions-in-batch-changes#permissions-in-batch-changes)

image

## After (locally)

image

## Why

The markdown was malformed in two ways:

1. There was a blank line between the header row and the `|---|` separator row.
   GFM requires the separator to be on the line immediately after the header,
   otherwise the whole block falls back to a paragraph.
2. The rows were missing leading/trailing `|` pipes, unlike every other table in
   the repo.

## How

Rewrote the table using the standard pipe-delimited form already used elsewhere
in the docs (e.g., `docs/tutorials/index.mdx`). No content changes — the actions
and ✅/❌ values are identical to the previous source.

## Test plan

- `pnpm run dev` and visit `/batch-changes/permissions-in-batch-changes`
- Confirm the "Batch Change access for each permission level" section renders
  as a real table with three columns (Batch change action / Read / Admin)
- Confirm no other pages are affected (only this one file changed)


enriquegh (approved):
Nice! LGTM via the preview deployment

PR #1781, opened 2026-06-10
docs: reframe k8s docker-in-docker executors around threat model

michaellzc:
Reframes the Docker-in-Docker (DinD) Kubernetes executor docs away from a "not recommended / requires privileged pod" warning toward a proper threat-model explanation. Adds a "Security considerations" section clarifying that running untrusted code is inherent to *all* executor runtimes (not unique to DinD) and that defense in depth bounds the blast radius regardless of runtime. Documents recommended hardening: dedicated/isolated node pools, gVisor sandboxing (e.g. GKE Sandbox), and network restrictions (blocking private ranges and the cloud metadata service). The goal is to help customers evaluate `privileged: true` in context rather than be scared off it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/self-hosted/executors/deploy-executors-dind.mdx
+++ b/docs/self-hosted/executors/deploy-executors-dind.mdx
@@ -1,18 +1,17 @@
 # Deploying Sourcegraph executors on Kubernetes (docker-in-docker)
 
 <Callout type="warning">
-	Docker-in-Docker Kubernetes executors are in beta and are not recommended for production use.
-	This deployment mode requires privileged pod access and does not use Firecracker isolation.
-	For production workloads, deploy using
+	Docker-in-Docker Kubernetes executors are in beta. For production workloads where you
+	want the strongest reliability and support guarantees, you can also deploy via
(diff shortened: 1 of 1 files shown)


docs/self-hosted/executors/deploy-executors-dind.mdx
```diff
@@ -21,12 +20,6 @@ Ensure you have the following tools installed:
 -   [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl)
 -   [Helm](https://helm.sh/) if you're installing Sourcegraph with `helm`.
 
-#### Deployment via Kustomize
-
-Please refer to the [Sourcegraph Kustomize docs](/self-hosted/deploy/kubernetes/kustomize) for the latest instructions.
-
-To include Executors dind, see [configure Sourcegraph with Kustomize](/self-hosted/deploy/kubernetes/configure) on how to specify the component.
-
```
michaellzc, line 29: intentional. let's not advertise it anymore

DaedalusG (approved):

PR #1790, opened 2026-06-22
PowerShell example in quickstart.mdx will change the registry data ty…

rvkasper:
…pe of Path - update it

The current example will change the underlying data type of the Path environment variable (in HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment) from REG_EXPAND_SZ to REG_SZ, which will have undesirable consequences down the line, as REG_EXPAND_SZ is the required data type if you wish to use unexpanded paths in the Path statement... such as %SystemRoot% 😮


rvkasper:
More background on the issue - MS is apparently not going to fix:
https://github.com/dotnet/runtime/issues/1442

PR #1796, opened 2026-06-29
docs: clarify AWS PrivateLink firewall requirements

eugenio-sanchez-sg:
Clarifies that AWS PrivateLink setup requires the network path behind the VPC endpoint service to allow inbound traffic from Sourcegraph-provided source CIDR ranges.

Adds an FAQ for cases where the VPC endpoint is accepted and available, but Sourcegraph requests still time out because backend firewall, NLB, security group, or NACL rules block traffic.


filiphaftek (approved):
TY

PR #1798, opened 2026-06-30
abc: initial docs

bobheadxi:
Starting point!

Closes https://linear.app/sourcegraph/issue/EPD2-158/bca-product-documentation


eseliger (approved):
thanks!

PR #1824, opened 2026-07-23
executor: add `executor install batch-exec` instructions

burmudar:
Companion fix to https://linear.app/sourcegraph/issue/SVC-2509/fix-single-binary-executor-install-to-include-batch-exec


marcleblanc2:
Hey @burmudar and @DaedalusG, is this doc change still accurate / needed?

PR #1827, opened 2026-07-29
Add supported countries page

eseliger:


valerybugakov (approved):

lilysch-sg (commented):
PR approved from Legal - following the supported regions policy approach that Anthropic and OpenAI use to reflect the regions we permit use of Sourcegraph in.

PR #1829, opened 2026-07-31
docs/observability: add Microsoft Teams alert notifier

taras-yemets:
Microsoft Teams users need clear setup guidance for routing Sourcegraph observability alerts to a channel without relying on retired legacy connectors. This adds a Microsoft Teams example alongside the existing notifier configurations, using a Teams Workflow, the `msteams` notifier type, and its complete signed URL. It also calls out that no separate token is required and that the URL must be treated as a secret.

Related product change: https://github.com/sourcegraph/sourcegraph/pull/14405

Linear: https://linear.app/sourcegraph/issue/SVC-2528/add-microsoft-teams-support-for-observability-alerts

### DIFF
--- a/docs/self-hosted/observability/alerting.mdx
+++ b/docs/self-hosted/observability/alerting.mdx
@@ -64,7 +64,7 @@ Once configured, Sourcegraph alerts will automatically be routed to the appropri
 
 #### Microsoft Teams
 
-Create a Microsoft Teams Workflow using [Post to a channel when a webhook request is received](https://support.microsoft.com/en-us/office/create-incoming-webhooks-with-workflows-for-microsoft-teams-8ae491c7-0394-4861-ba59-055e33f75498), then copy its signed Workflow URL.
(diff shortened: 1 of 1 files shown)


docs/self-hosted/observability/alerting.mdx
```diff
@@ -62,6 +62,25 @@ Once configured, Sourcegraph alerts will automatically be routed to the appropri
 
 > NOTE: Learn more about generating a [Slack incoming webhook URL](https://api.slack.com/messaging/webhooks)
 
+#### Microsoft Teams
+
+Create a Microsoft Teams Workflow using [Post to a channel when a webhook request is received](https://support.microsoft.com/en-us/office/create-incoming-webhooks-with-workflows-for-microsoft-teams-8ae491c7-0394-4861-ba59-055e33f75498), then copy its signed Workflow URL.
```
burmudar: Did we test this with an actual teams channel?

burmudar (approved):

docs/self-hosted/observability/alerting.mdx
taras-yemets: Yes, this was tested end to end against a real Microsoft Teams channel. I added the site configuration, synthetic mutation, and observed delivery results to the test plan in sourcegraph/sourcegraph#14405. I also updated this guide to use Microsoft’s current generic **Send webhook alerts to a channel** template and clarify the unauthenticated webhook requirement.

PR #1838, opened 2026-08-20
Remove Code Finder MCP beta badge

janhartman:
GA launch is today, so the beta badge should be removed.


burmudar (approved):
Drive by

Image

PR #1849, opened 2026-08-31
Keep legacy selectors connected to current versions

HawkSE:
## Summary

- read the canonical version manifest through the stable `main` Vercel origin, avoiding Cloudflare challenges on cross-subdomain browser requests
- keep the full version list on current docs while arranging legacy menus as latest, selected, then older versions
- document the manifest origin and legacy filtering behavior in the publishing skill

## Test plan

- `npx --yes pnpm@10.25.0 run build`
- browser-verified the Vercel manifest returns CORS JSON from a legacy origin
- browser-verified current docs retain the full list

### DIFF
--- a/src/components/VersionSelector.tsx
+++ b/src/components/VersionSelector.tsx
@@ -14,7 +14,7 @@ import {Fragment, useEffect, useState} from 'react';
 
 const versionsUrl =
 	process.env.NEXT_PUBLIC_DOCS_VERSIONS_URL ??
-	'https://sourcegraph-docs-git-main-sourcegraph-f8c71130.vercel.app/docs/api/versions';
+	'https://sourcegraph.com/docs/api/versions';
 
 function isVersion(value: unknown): value is VersionI {
 	if (typeof value !== 'object' || value === null) return false;
--- a/src/components/VersionSelector.tsx
+++ b/src/components/VersionSelector.tsx
@@ -14,7 +14,7 @@ import {Fragment, useEffect, useState} from 'react';
 
 const versionsUrl =
 	process.env.NEXT_PUBLIC_DOCS_VERSIONS_URL ??
-	'https://sourcegraph.com/docs/api/versions';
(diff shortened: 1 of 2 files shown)


src/components/VersionSelector.tsx
```diff
@@ -14,7 +14,7 @@ import {Fragment, useEffect, useState} from 'react';
 
 const versionsUrl =
 	process.env.NEXT_PUBLIC_DOCS_VERSIONS_URL ??
-	'https://sourcegraph.com/docs/api/versions';
+	'https://sourcegraph-docs-git-main-sourcegraph-f8c71130.vercel.app/docs/api/versions';
```
enriquegh: hmm is this URL safe to use? Will it always point to main? 

Can we see if the Security team has any advice on how to accomplish this? If we must, I think a better URL would be https://sourcegraph-docs-v2.vercel.app/docs as that seems to be static to our main deployment but I'd prefer if we kept this to sourcegraph.com domain

enriquegh (changes requested):

valerybugakov (approved):

src/components/VersionSelector.tsx
HawkSE: I think the only way to get around this would be to add some sort of Cloudflare WAF exception for sourcegraph.com/docs/api/versions to keep this all in-tact.

HawkSE: With the latest change to Cloudflare bot detection from the SEC team, this is no longer failing and the workaround is no longer needed. I've restored the Sourcegraph.com domain for this update.

valerybugakov (approved):

HawkSE:
This version of the changes were meant to correct some issues with the live site. It turns out the Cloudflare issues were preventing the previous PR from propagating to all the locations. The current live version of the site is the ideal version of my changes and this PR and associated branches are no longer needed.

PR #1851, opened 2026-09-03
Point stale docs.sourcegraph.com links at sourcegraph.com/docs

bahrmichael:
`docs.sourcegraph.com` is no longer the docs host. This rewrites the remaining absolute references to their `sourcegraph.com/docs` equivalents.

The rewrite is purely mechanical — only the host portion changes (`docs.sourcegraph.com/<path>` becomes `sourcegraph.com/docs/<path>`). Paths that were reorganised since (for example `admin/observability/tracing` and `batch_changes/references/troubleshooting`) are already covered by permanent redirects in the docs site's redirect map, so the rewritten links resolve.

In `src-cli`, `CHANGELOG.md` is deliberately left alone: those entries are a historical record. In the docs repo the change is scoped to `docs/cli/references/**`, which is generated from `src-cli` by `src doc` — the companion `src-cli` changeset fixes the generator's source strings so the docs stay correct after the next regeneration.

[_Created by a Sourcegraph agentic batch change._](https://sourcegraph.sourcegraph.com/batch-change-agents/453)


marcleblanc2:
Hey @bahrmichael, these pages come from the src-cli doc generator script, and the links should be fixed in the src-cli repo, so the next time the doc generator script runs, it updates these files.

PR #1858, opened 2026-09-06
ci/links: Add PR check for broken links

marcleblanc2:
Linear [FE-499: Fix doc site issues](https://linear.app/sourcegraph/issue/FE-499/fix-doc-site-issues)

## Problem

- Our docs site has hundreds of broken links
- `dev/check-links.mjs` finds broken internal links and anchors, but it isn't run automatically, so PRs can easily break links (renaming a heading, moving or deleting a page) without anyone noticing

## Solution

- Updated the script to also work as a PR check, with additional functions beyond what's run when used as a CI test in Vercel builds
- PR check to run the script and report if the PR breaks links
- It runs the script (with `--check-anchors`) on both the PR head and its merge base, and diffs the findings
- This catches both directions:

- **Outbound**: a changed page links to a page or `#heading` that doesn't exist
- **Inbound**: the PR renames a heading or removes/moves a page that other, unchanged pages link to — those show up as findings in files the PR didn't touch

- Pre-existing broken links are ignored by the PR check
- The comment is created / updated in place, and once the PR is fixed, the PR check passes and the comment is updated to say so
- A PR that never broke anything gets no comment

## Verification

PR check comment in test PR: https://github.com/sourcegraph/docs/pull/1895#issuecomment-5611164928

### Broken links found

Screenshot 2026-09-09 at 20 05 31

### Broken links fixed

Screenshot 2026-09-09 at 20 06 31

## Absolute self-links and external links

- Absolute links to this site (`https://sourcegraph.com/docs/…`, `http://…`, `//…`, `www.`, the legacy `https://docs.sourcegraph.com/…`) fail the check even when the target exists: they leave the Vercel preview and local dev, and hide moved pages behind redirects. The finding names the relative link, following `src/data/redirects.ts` when the page moved. Version-pinned links (`/@5.1/…`) stay external
- External links on lines this PR added are requested (HEAD, then GET on an error status, following redirects); only 404 and 410 are findings, so rate limits, bot blocks, 5xx and network errors never fail a PR. Placeholder hosts (`*.example.com`, `localhost`, templated `<host>`) are skipped
- Findings with a fix become one suggested-change review comment per line, which the author can apply from the PR. Suggestions already on the PR are not posted again
- #1899 clears the 67 existing absolute self-links so this check starts from zero

Test PR: #1900 (report comment + one review suggestion; the `#sampling` anchor deliberately does not exist, so that link gets no suggestion; a second run posted nothing new)

## Related

- Draft PR #1562 proposes a daily Slack digest with a separate reimplementation of this script
  - Instead, this PR improves on the existing script, and gates PRs
- PR https://github.com/sourcegraph/docs/pull/1860 enabled external link checkers to find broken links again

## Amp threads

- [Broken link PR check](https://ampcode.com/threads/T-01a0753f-0f4f-7478-b36c-87466e7c0261)
- [Asset case mismatch](https://ampcode.com/threads/T-01a07597-43c0-751b-8c49-6e5809e714d2)
- [Docs - Fix broken heading links](https://ampcode.com/threads/T-01a07623-9d65-7356-96b8-2bebb31ffa5a)
- [Self-links and external links](https://ampcode.com/threads/T-01a08a01-44c1-775b-84d0-d67ff9501905)

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -70,8 +70,11 @@ jobs:
                   existing_comment=$(gh api "repos/$GITHUB_REPOSITORY/issues/$PR_NUMBER/comments" \
                       --paginate --jq ".[] | select(.body | startswith(\"$marker\")) | .id" | head -n 1)
 
-                  # Only comment when there is something to report, or a stale report to resolve
-                  if [ -z "$existing_comment" ] && [ "$BROKEN" != true ]; then
+                  # A clean PR carries no comment: remove a stale report once fixed
+                  if [ "$BROKEN" != true ]; then
+                      if [ -n "$existing_comment" ]; then
+                          gh api --method DELETE "repos/$GITHUB_REPOSITORY/issues/comments/$existing_comment"
+                      fi
                       exit 0
                   fi
 
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -70,15 +70,16 @@ jobs:
                   existing_comment=$(gh api "repos/$GITHUB_REPOSITORY/issues/$PR_NUMBER/comments" \
                       --paginate --jq ".[] | select(.body | startswith(\"$marker\")) | .id" | head -n 1)
 
-                  # A clean PR carries no comment: remove a stale report once fixed
-                  if [ "$BROKEN" != true ]; then
-                      if [ -n "$existing_comment" ]; then
-                          gh api --method DELETE "repos/$GITHUB_REPOSITORY/issues/comments/$existing_comment"
-                      fi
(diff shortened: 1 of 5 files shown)


pjlast (approved):
Approving as is (it's great!), but I do wonder if there are cases where we might link to an external page that requires some sort of authorization, which could return a 404 and trigger the check unnecessarily 🤔

marcleblanc2:
> Approving as is (it's great!), but I do wonder if there are cases where we might link to an external page that requires some sort of authorization, which could return a 404 and trigger the check unnecessarily 🤔

Those should return a 401 / 403 unauthorized instead of 404? I guess we'll find out.

PR #1879, opened 2026-09-08
feat/redirects: Apply heading (#fragment) redirects in the browser

marcleblanc2:
## Problem

`src/data/redirects.ts` contains 386 entries whose `source` includes a `#fragment`, e.g.

```ts
{source: '/cody/quickstart#introduction', destination: '/cody/quickstart#cody-quickstart'}
```

Browsers never send the fragment to the server, so `src/middleware.ts` (which matches `r.source === pathname`) has never fired for any of them — and neither did the Next `redirects()` config they were originally added to in 2023. Verified on the live site: https://sourcegraph.com/docs/batch_changes/explanations/permissions_in_batch_changes#code-host-interactions-in-batch-changes serves the not-found page with no `Location` header.

This also means there has been no way to add a redirect when renaming a heading, so inbound anchored links silently break.

## Change

- `src/components/FragmentRedirect.tsx`: client component rendered once in the root layout. After load (and on `hashchange`) it looks up `pathname + location.hash` in the `#` entries of the same redirects list. Same-page hit → `history.replaceState` + scroll to the new heading (no reload). Cross-page hit → `location.replace`. It also handles the two-step case where the middleware redirects the path and the browser carries a stale fragment onto the new page.
- `src/data/redirects.ts`: exported as ESM (`export const updatedRedirectsData`) so the middleware and the client component import the same list. No entries were moved, added, or removed. Added a header comment explaining which entries are handled where.
- `src/middleware.ts`: `require` → `import`. Behaviour unchanged.

Heading renames are now handled by adding one line to the existing list: `{source: '/page#old-heading', destination: '/page#new-heading'}`.

## Verification (dev server, headless Chromium)

| Requested URL | Landed on |
|---|---|
| `/batch_changes/explanations/permissions_in_batch_changes#code-host-interactions-in-batch-changes` (404 today) | `/batch-changes/permissions-in-batch-changes#code-host-interactions-in-batch-changes`, scrolled to heading |
| `/cody/quickstart#introduction` (same-page rename) | `/cody/quickstart#cody-quickstart`, no reload, scrolled to heading |
| `/cody/clients/install-vscode#requirements` (what the browser lands on after the middleware's 307 from `/cody/overview/install-vscode#requirements`) | `/cody/clients/install-vscode#prerequisites` |
| `/cody/quickstart#prerequisites` (valid anchor, control) | unchanged |
| `/admin/http_https_configuration` (middleware path redirect, control) | 307 → `/self-hosted/http-https-configuration` |

`eslint` clean on changed files; `node dev/check-links.mjs` passes; `tsc` errors are pre-existing (`contentlayer/generated`).

## Notes / follow-ups (not in this PR)

- The `#` entries now ship to the client as part of the redirects chunk (~15 KB compressed, cached across pages). Most of the 386 point at long-gone Cody pages and could be pruned.
- The bare `/batch_changes/explanations/permissions_in_batch_changes` (no anchor) still 404s; it needs a fragment-less entry.
- 203 duplicate sources and 108 entries whose destination no longer exists were found while auditing; separate cleanup.
- Pre-existing: `createRedirectUrl` hardcodes `/docs`, so middleware redirects 404 in local dev; `notFound()` pages return HTTP 200 in production.

## Amp threads

- [Redirect list cleanup](https://ampcode.com/threads/T-01a07e6f-db41-74af-bbeb-f8952e637289)

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -1,4 +1,5 @@
 import {Providers} from '@/app/providers';
+import {FragmentRedirect} from '@/components/FragmentRedirect';
 import {Layout} from '@/components/Layout';
 import {TECHNICAL_CHANGELOG_RSS_URL} from '@/data/constants';
 import clsx from 'clsx';
@@ -89,6 +90,7 @@ export default function RootLayout({children}: {children: React.ReactNode}) {
 				className="flex min-h-full bg-light-bg dark:bg-dark-bg"
 			>
 				<Suspense fallback={null}>
+					<FragmentRedirect />
 					<Providers>
 						<Layout>{children}</Layout>
 					</Providers>
--- a/src/components/FragmentRedirect.tsx
+++ b/src/components/FragmentRedirect.tsx
@@ -1,0 +1,94 @@
+'use client';
+
+import {updatedRedirectsData} from '@/data/redirects';
+import {usePathname} from 'next/navigation';
+import {useEffect} from 'react';
+
+const basePath = process.env.NEXT_PUBLIC_DOCS_BASE_PATH || '';
+
+function splitHash(url: string): {path: string; hash: string} {
+	const i = url.indexOf('#');
+	return i === -1
+		? {path: url, hash: ''}
+		: {path: url.slice(0, i), hash: url.slice(i)};
+}
+
+// Entries in src/data/redirects.ts whose source has a `#fragment`. The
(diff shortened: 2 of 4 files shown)


marcleblanc2:
Moved to #1934 (branch renamed to `marc/fragment-redirects`).

PR #1887, opened 2026-09-09
feat/site: Add navigation links on 404 page

marcleblanc2:
## What

The 404 page only offered "Go back home". It now also shows, when applicable:

- **Go back to `<path>`** — the docs page the user was on before this one
- **Go up to `<path>`** — the closest ancestor of the requested URL that is a real docs page (preview pages excluded). Omitted when that ancestor is the root (the home link covers it) or is the same page as the back link.

## Before

- Only see a clickable link to the home page, which is almost never where I want to go

image

## After

- Also see a link back to the previous page I was on
- If there's also a valid parent page, which is different than the previous page, then show that as an option as well (useful because we have so many broken links / redirects to old subpages, and the content's likely been moved to the parent page, or another page I can click to from the parent page)

image

### DIFF
--- a/src/app/not-found.tsx
+++ b/src/app/not-found.tsx
@@ -1,4 +1,8 @@
-import Link from 'next/link';
+import {NotFoundLinks} from '@/components/NotFoundLinks';
+import {allPosts} from 'contentlayer/generated';
+
+// Preview pages 404 without ?preview, so they are not valid targets.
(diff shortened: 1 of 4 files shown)


pjlast (approved):
My immediate reaction is "Why not just click on the back button in the browser?" 🤔 

But I do like the closest ancestor link as a way of saying "This page doesn't exist, but maybe info on it can be found here"

PR #1890, opened 2026-09-09
feat/dev: Add Cloudflare traffic reports

marcleblanc2:
Adds `viewership-metrics/page-views-report.mjs`, a dependency-free Node script that pulls 90 days of Cloudflare analytics for `/docs`, `/changelog` and `/blog` on sourcegraph.com and writes Markdown reports to `viewership-metrics/reports/` (gitignored). See `viewership-metrics/README.md`.

## What it reports

- Page views (HTML 200s), visits, 3xx, 404 and 5xx per path, in four files sorted by path, requests, redirects and errors
- `redirect-rules.md`: hits per rule in `src/data/redirects.ts`, flagging rules shadowed by an earlier duplicate source and redirect chains (a destination that is itself another rule's source)

## Filters

Bot Management `likely_human`, excluding ASN `Hetzner Online GmbH` and country `CN`. Skips `/_next`, static assets, scanner probe paths and trailing-slash redirects.

## Findings from the first 90-day run

- 3,041 paths, 524k requests, 65k redirects, 16k 404s, 2.8k 5xx
- 1,324 redirect rules: 962 live, 362 shadowed (never match), 689 live with zero hits
- 269 live rules chain into another rule; longest chain is 5 extra hops. The `precise_code_navigation` rules alone send ~5,500 visitors through 2 or 3 redirects
- `/changelog/.api/changelog.v1.ChangelogService/ListReleasePosts` returned 1,880 5xx
- `/blog/feed.atom` (3,410) and `/blog/feed.xml` (1,000) 404 while `/blog/rss.xml` redirects 21,790 times
- The docs site returns 200 for unknown paths, so deleted pages and probes like `/docs/.zshrc` count as views

## Test

```sh
CLOUDFLARE_API_TOKEN=... npm run page-views-report -- --days 7
ls viewership-metrics/reports/
```

## Amp threads

- [Cloudflare docs site usage metrics](https://ampcode.com/threads/T-01a08479-a4c8-72ad-8f31-fe7ecc43bf34)
- [Run page-views-report against Cloudflare](https://ampcode.com/threads/T-01a084d0-e1fb-77e7-b6fc-2897beec8ef6)

PR #1897, opened 2026-09-10
dev/viewership-metrics: Probe every redirect rule on the live site

marcleblanc2:
Stacked on #1890.

`npm run probe-redirects` requests every rule in `src/data/redirects.ts` on sourcegraph.com, follows the redirects like a browser, and writes `viewership-metrics/reports/redirect-probe.json`: per rule, every hop, final URL/status, whether the first redirect is the one the rule promises, whether the `#fragment` the user ends up with exists on the page, and the Cloudflare rows for source, destination and final page.

Headline numbers from the first run (962 live rules):

- 735 redirect as written, 195 redirect somewhere else, 32 do not redirect at all
- 224 end on a 404; 6 loop forever (`/graphbook*` → `sourcegraph.com/handbook` redirects to itself; `/@v5.3.0/admin/observability/*` ping-pongs with `docs.sourcegraph.com`)
- 230 live rules have a `#fragment` in the source. Browsers never send fragments, so none of them can match as written; 200 fall through to a bare-path rule, 30 to the page itself
- 22 rules put a `#fragment` in the destination; the anchor exists on the final page for 9

See `viewership-metrics/README.md` → Redirect probe for the field meanings.


marcleblanc2:
Moved to #1939 (branch renamed to `marc/redirect-probe`).

PR #1908, opened 2026-09-11
Remove duplicated docs build check executions

marcleblanc2:
This is task 6a of the Vercel audit tracked in #1905.

## Problem

- `next build` evaluated the asynchronous Next.js config twice before printing the Next.js banner

- Logs recorded `phase-production-build` twice in process 31244, and both evaluations ran all four checks each

## Changes

- Run the four checks in order from the `build` package script, before `next build`. Vercel invokes this script once, while Next.js may evaluate its config more than once.

- Chain commands with `&&`, so any failed check exits non-zero and prevents the deploy build from continuing

- Keep the existing `generate-mermaid-logos` script name for compatibility and point it at the real `dev/generate-mermaid-icons.mjs` file

- Update the build-command guidance to distinguish `npm run build` from direct `next build`

## Test plan

```sh
mise exec -- pnpm install --frozen-lockfile
mise exec -- pnpm build
mise exec -- pnpm dev --port 3106
curl http://localhost:3106/
mise exec -- npx tsc --noEmit
```

Note: #1907 also edits `next.config.js`; whichever merges second needs a trivial rebase.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,13 +8,13 @@
 	},
 	"scripts": {
 		"dev": "next dev",
-		"build": "next build",
+		"build": "node dev/check-links.mjs && node dev/check-filenames.mjs && node dev/check-images.mjs && node dev/generate-mermaid-icons.mjs && next build",
 		"start": "next start",
 		"lint": "next lint",
 		"check-links": "node dev/check-links.mjs",
 		"check-filenames": "node dev/check-filenames.mjs",
 		"check-images": "node dev/check-images.mjs",
-		"generate-mermaid-logos": "node dev/generate-aws-icons.mjs",
+		"generate-mermaid-logos": "node dev/generate-mermaid-icons.mjs",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -4,13 +4,13 @@
 	"private": true,
 	"scripts": {
 		"dev": "next dev",
-		"build": "next build",
+		"build": "node dev/check-links.mjs && node dev/check-filenames.mjs && node dev/check-images.mjs && node dev/generate-mermaid-icons.mjs && next build",
```
burmudar: there must be an aggregate way to have all of these defined in one checks.mjs file? 

`node dev/checks.mjs` - default everything
`node dev/checks.mjs links`
`node dev/checks.mjs filenames`

marcleblanc2: Good thinking. I've had Amp write one script which calls the other 3.

burmudar (approved):

PR #1880, opened 2026-09-08
ci/redirects: Add Redirect check for PRs

marcleblanc2:
Linear [FE-499: Fix doc site issues](https://linear.app/sourcegraph/issue/FE-499/fix-doc-site-issues)

Stacked on #1858 (base branch `check-links-pr-check`); merge that one first.

## Problem

- `src/data/redirects.ts` has ~900 entries that don't work: destination pages that no longer exist, `#fragments` in the source (browsers never send them, so the entry can never match), duplicates of an earlier source, chains to other redirects, `/docs/` prefixes
- Nothing checks the file, so PRs keep adding broken redirects, and PRs that move a page or rename a heading silently break the redirects pointing at it

## Solution

- `dev/check-redirects.mjs` validates every entry against the pages and headings in `docs/`
- "Check redirects" PR check runs the script on the PR head and its merge base, and reports only what the PR introduced; the ~900 pre-existing problems are ignored
- Runs in parallel with "Check links"
- The comment groups findings by problem, and each section says how to fix its entries, so authors learn how to write a correct redirect:
  - Source overshadows a docs page that exists
  - Source has a `#fragment`, so this redirect can never match
  - Source or destination starts with `/docs`
  - Source already has an earlier entry, so this one is never used
  - Destination is another redirect (shows the final destination)
  - Destination page does not exist
  - Destination heading does not exist
- Line numbers link to the entry in `redirects.ts` on the PR branch
- The comment is created / updated in place, and once the PR is fixed, the PR check passes and the comment is updated to say so
- A PR that never broke anything gets no comment
- `pnpm check-redirects` runs it locally

## Verification

PR check comment in test PR: https://github.com/sourcegraph/docs/pull/1896#issuecomment-5612348499

Check redirects PR comment listing 8 broken redirects grouped by problem

## Related

- #1858 adds the "Check links" PR check this is built on

## Amp threads

- [Redirect list cleanup](https://ampcode.com/threads/T-01a07e6f-db41-74af-bbeb-f8952e637289)
- [Redirect check PR comment](https://ampcode.com/threads/T-01a088d9-b8fd-76fd-ba93-3a416c5829a3)

## Merge order for the PR-check stack

Trial-merged onto `main` in this order with no conflicts:

1. #1918 Vercel build log comment — independent; first so the other PRs' Vercel failures get a readable log
2. #1916 check-links report format — adds `dev/sync-review-comments.sh`, which #1880 calls
3. #1880 redirect check — needs #1916 merged first
[...]

PR #1913, opened 2026-09-11
test: PR checks on a broken PR, left broken (do not merge)

marcleblanc2:
Test PR to exercise the spell check, check-links, and check-redirects PR checks on a commit that breaks all three. This PR stays broken; #1914 applies the checks' suggested fixes and then fixes the rest by hand, for a side-by-side comparison.

The redirect check is not on `main` yet; the first commit copies `.github/workflows/check-redirects.yml` and `dev/check-redirects.mjs` from #1880 so it runs here.

Do not merge.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -97,18 +97,28 @@ jobs:
                   fi
 
             - name: Suggest fixes as review comments
-              # One suggested change per added line with a fix. Suggestions already on
-              # the PR (same file, line, and text) are not posted again.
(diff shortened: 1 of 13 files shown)


marcleblanc2:
Replaced by https://github.com/sourcegraph/docs/pull/1920, rebuilt on current main and the latest check code.

marcleblanc2:
Replaced by https://github.com/sourcegraph/docs/pull/1920, rebuilt on current main and the latest check code.

PR #1914, opened 2026-09-11
test: PR checks on a broken PR, then remediated (do not merge)

marcleblanc2:
Same commit as #1913, which stays broken for comparison. Here, a second commit applies the fixes the checks suggested (check-links and spell check review suggestions), and a third fixes what the checks could only report: the dead external link, missing page/heading/wrong-case links, the inbound anchor link, the spelling suggestion CSpell got wrong, and the broken redirects.

The redirect check is not on `main` yet; the first commit copies `.github/workflows/check-redirects.yml` and `dev/check-redirects.mjs` from #1880 so it runs here.

Do not merge.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -97,18 +97,28 @@ jobs:
                   fi
 
             - name: Suggest fixes as review comments
-              # One suggested change per added line with a fix. Suggestions already on
-              # the PR (same file, line, and text) are not posted again.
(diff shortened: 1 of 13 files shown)


marcleblanc2:
Replaced by , rebuilt on current main and the latest check code.

marcleblanc2:
Replaced by https://github.com/sourcegraph/docs/pull/1921, rebuilt on current main and the latest check code.

PR #1953, opened 2026-09-11
check-links: one suggestion per fix, synced with findings; one fact per line in reports

marcleblanc2:
Follow-ups from testing the PR checks on #1913 / #1914.

- Review comments: one suggested change per finding (not one per line), no review body. Each comment starts with an HTML marker; the workflow deletes suggestions whose finding is gone (or that GitHub could no longer place, `line: null`) and skips ones already posted, so resolved suggestions disappear like the spell check's do.
- Summary comment and review comments put line, link, problem, and fix each on their own line.
- Absolute links to this site get their own **Absolute links** section instead of being lumped into Outbound.
- Case-mismatch findings now come with a fix (`/Code-Search/queries` → `/code-search/queries`).
- Wording: "Write links on this site as relative paths", "fix the inbound links on these other pages", dropped `https://docs.sourcegraph.com/…`; reproduce command matches `package.json` (`pnpm check links …`).

Tested locally against the `test-pr-checks-broken` branch with the CI recipe (baseline from `origin/main`, `--diff`, `--review`); build-mode run (`node dev/check-links.mjs`) still clean.

Trade-off: when several fixes sit on one line, applying one suggestion outdates the others until the next run re-posts them, because GitHub will not batch overlapping suggestions.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -97,18 +97,28 @@ jobs:
                   fi
 
             - name: Suggest fixes as review comments
-              # One suggested change per added line with a fix. Suggestions already on
-              # the PR (same file, line, and text) are not posted again.
-              if: steps.check.outputs.broken == 'true' && github.event.pull_request.head.repo.full_name == github.repository
+              # One suggested change per finding with a fix. Suggestions already on the
+              # PR (same file, line, and link) are not posted again; suggestions for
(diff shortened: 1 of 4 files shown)


marcleblanc2:
Closed: #1916 was already approved, so I renamed its branch back and reopened it instead of using this replacement.

PR #1977, opened 2026-09-17
Serve site at a consistent, configurable basePath, for all deployments

marcleblanc2:
## Problem

- The doc site has moved a few times over the years, and will move again in the coming months

- The whiplash of these moves has resulted in a hot mess of adaptations

- Prod serves the site at `sourcegraph.com/docs`

- Preview and dev deployments have served it at either `/` or `/docs` at different times in history, most recently `/`

- Root-relative URLs which Next.js does not prefix (``, `<a href>`, `fetch()`) therefore worked in preview but broke in prod

## Change

- This PR creates a consistent pattern: 
  - There is now exactly one (1) config for `DOCS_PROD_ORIGIN`, in `docs.config.js`, set to `https://sourcegraph.com`
  - There is now exactly one (1) config for `DOCS_BASE_PATH`, in `docs.config.js`, set to `/docs`
  - Standardized paths using `withBasePath(url)` from `src/lib/utils.ts`

- Every deployment is now served under `DOCS_BASE_PATH`

- `/` redirects to `DOCS_BASE_PATH`, so preview URLs and `localhost:3000` land on the site automagically

- `/v/<latest>/x` redirects to `/DOCS_BASE_PATH/x` instead of prod, so previews can be tested end to end

- The many different adaptations to conjure these have been centralized on these configs

- Updated dev tooling and docs to match

## Verification

- `tsc` and `eslint` clean
- Against `next dev`:

| Request | Result |
|---|---|
| `/` | 307 → `/docs` |
| `/docs/self-hosted/deploy` | 200, every `` starts with `/docs/` |
| `/docs/self-hosted/deploy.md` | 200 |
| `/docs/v/6.0/cody` | 307 → `https://6.0.sourcegraph.com/cody` |
| `/docs/v/7.7/cody` | 307 → `<same host>/docs/cody` |
| `/docs/integration/img/disable_extension.png` | 307 → `/docs/integration/img/disable-extension.png` |

On `main`, the same local setup renders `src="/sourcegraph-mark.svg"` and it 404s: this bug would have been caught on a preview.


marcleblanc2:

The crawl confirms the three broken-image sources this PR fixes (`/sourcegraph-mark.svg` via `MobileNavigation.tsx` and the LinkCard `imgSrc` on 67 pages, `/aws-ami-marketplace.png`, `/postgres_collation_affected_customers.png`). Two related findings it does not cover:

1. `docs/integration/gitlab.mdx:29` references `img/gitlab-code-intel.gif`. Nothing serves `docs/**/img` (`curl https://sourcegraph.com/docs/integration/img/disable-extension.png` → 404), and it is the only MDX reference into that folder; the other ~50 files in `docs/integration/img/` are orphaned. Moving the gif into `public/` would conflict with this PR, so it may be easiest to fold in here.
2. The canonical: this PR makes it `config.DOCS_BASE_PATH` (still site-wide), so every page keeps telling search engines it is a duplicate of `/docs`. Per-page canonical is in #2001; expect a trivial conflict on `layout.tsx` if both land.

enriquegh (approved):
Looks good! Do we need to add any of this to the old versions of our docs (e.g. 7.7.sourcegraph.com, etc)?

That shouldn't block though so approving

marcleblanc2:
> Do we need to add any of this to the old versions of our docs (e.g. 7.7.sourcegraph.com, etc)?

@enriquegh Amp's response below, and I'm trying to make heads or tails of it. 

Do we have customers or staff who flip between different versions of the site?

> - Not for old versions, but future versions will include this change
> - The versioned sites (7.7.sourcegraph.com etc.) are frozen Vercel deployments served at the root of their own subdomain, with no basePath, so root-relative URLs already resolve there (`https://7.7.sourcegraph.com/sourcegraph-mark.svg` is a 200 today)
> - The bugs only show up when the site is served under `/docs`
> - Applying this PR to an old version would move it to `<old-version>.sourcegraph.com/docs/...`, breaking existing links, and requiring updating the `/v/<version>/*` redirects

PR #1976, opened 2026-09-17
perf: prefetch sidebar links on hover, not in viewport

marcleblanc2:
## Problem
Next prefetches every `<Link>` that enters the viewport. The sidebar shows dozens at once, so a single page view of `/docs/batch-changes/delete-a-batch-change` fetched 84 RSC payloads: 377 KB on the wire, 2.4 MB decoded, most of which the reader never clicks.

## Fix
New `NavigationLink` wraps `next/link` with `prefetch={hasIntent ? null : false}`, flipping to the default prefetch on `onMouseEnter`/`onFocus`. This is the [hover-triggered prefetch pattern from the Next docs](https://nextjs.org/docs/app/guides/prefetching#hover-triggered-prefetch). `Navigation.tsx` swaps `Link` for it at its three link sites; nothing else changes.

## Tradeoffs
- First click on a link the user never hovered (touch devices) is a normal, non-prefetched navigation. Static pages, so still fast.
- `prefetch={false}` everywhere was the simpler option but also kills hover prefetch; this keeps it.
- `origin/enrique/sidebar-navigation` touches `Navigation.tsx`; expect a small merge conflict at the import and the three `<Link>` tags.

## Verification
`tsc`, `pnpm lint`, `pnpm build` clean. On the preview, open DevTools → Network, filter `?_rsc=`: the page load should fetch a handful, not ~80; hovering a sidebar link should fetch its payload.

PR #1978, opened 2026-09-17
fix: do not prefetch external sidebar links

marcleblanc2:
Stacked on #1976 (`marc/perf/sidebar-prefetch`); retarget to `main` once that merges.

## Problem
`navigation.ts` has Pricing and Changelog entries with `href: 'https://sourcegraph.com/...'`. `next/link` treats an absolute URL on the current origin as an app route, so every page load prefetched the marketing pricing page (65 KB of HTML in the HAR). On versioned docs, `prependVersion` also turns them into `/v/<version>/https://sourcegraph.com/pricing`.

## Fix
- `NavigationLink` renders a plain `<a>` for any `http(s)://` href.
- `prependVersion` only prefixes paths that start with `/`.

## Verification
`tsc`, `pnpm lint`, `pnpm build` clean. On the preview, the Network tab should show no request to `sourcegraph.com/pricing` on page load, and the Pricing link should still navigate.


marcleblanc2:

Same root cause, one more instance outside this PR's files: the Changelog link in the desktop header (`src/components/Layout.tsx:78`) and mobile navigation (`src/components/MobileNavigation.tsx:111`) is also a `next/link` to `https://sourcegraph.com/changelog`. In the crawl that prefetch cost about 61 requests / ~500 KB per docs page and produced 1,155 "resource was preloaded but not used" console warnings. #2002 applies the same plain-`<a>` treatment to those two links.

PR #2006, opened 2026-09-17
Update docs latest version to 8.0

aditiabhang:
## Summary
- Set DOCS_LATEST_VERSION to 8.0
- Add Sourcegraph 7.7 to the version selector
- Update the legacy versions page

## Test plan
- `pnpm run check`
- `npx tsc --noEmit`
- Rendered `/legacy` locally and verified the expanded Sourcegraph 7.X list shows 7.7 through 7.0 in descending order


src/data/versions.ts
```diff
@@ -11,7 +11,11 @@ export const versions: VersionI[] = [
 	{
 		name: `v${config.DOCS_LATEST_VERSION}`,
 		label: 'latest',
-		url: '/docs'
+		url: 'https://sourcegraph.com/docs'
```
enriquegh: I think this should stay as /docs otherwise things break when running on dev/Amp orbs

enriquegh (changes requested):

src/data/versions.ts
aditiabhang: Of course. Not sure though how/why amp skill made this url change.

aditiabhang (commented):
Comment resolved

aditiabhang (commented):
Comment resolved.

aditiabhang (commented):
Url reverted to /docs instead of sourcegraph.com/docs

enriquegh (approved):

PR #2012, opened 2026-09-18
dev: Propose automating the monthly docs version bump

marcleblanc2:
- Amp generated, I haven't read through this yet

## Summary

Proposal to automate the monthly "Update docs latest version to X.Y" work, written up in `dev/versioning-proposal.md`. No behaviour change in this PR; it is a draft for discussion.

Today each minor takes five hand steps across four systems (infrastructure DNS PR behind an Entitle grant, cut and push a legacy branch, set its metadata, bump three files in this repo, map the domain in the Vercel dashboard). The three files here carry the same fact three times, the cut commit has in practice been the parent of the bump commit, and the release registry already knows which minors exist.

Proposed, in shippable order:

1. One `DOCS_VERSIONS` list in `docs.config.js`; `src/data/versions.ts` and `docs/legacy.mdx` derive from it. A frozen branch then needs no metadata commit.
2. A `push` to `main` workflow archives the previous version: push `HEAD^` to `docs-legacy-versions` as `vP_Q`, map `P.Q.sourcegraph.com` through the Vercel API, verify, report to `#alerts-vercel-doc-site`.
3. A daily workflow opens the bump PR when the release registry's latest minor differs from `DOCS_VERSIONS[0]`.
4. Pre-provision a whole major in `dns/sourcegraph.vercel.tf` once, instead of one minor per month.

Net: monthly work becomes reading and merging one bot-opened PR; the Notion runbook and the `publishing-docs-versions` skill shrink accordingly.

The list stays in the repo rather than being fetched at build time because `/v1/releases/sourcegraph` caps at the newest 50 releases with no paging (checked `limit`, `page`, `offset`), so it would lose old minors over time, and a frozen host has to pin its own version anyway.

Open questions are at the bottom of the doc: selector scope (current and previous major vs today's 14 entries), whether "parent of the bump commit" is an acceptable default cut rule, and who owns the two new secrets.

## Test plan

- Prose only. `prettier --check`, `markdownlint-cli2` and `cspell` clean on both files.
- Facts checked against `git log` in this repo, the `v7_7` branch in `docs-legacy-versions`, `dns/sourcegraph.vercel.tf` in `infrastructure`, the release registry API, and `https://7.7.sourcegraph.com/api/versions`.


dev/versioning-proposal.md
```diff
@@ -0,0 +1,134 @@
...
+   `vP_Q` of the `docs-legacy-versions` project.
+
+`git log -- docs.config.js` shows this ran five times since June 2026
+(#1782, #1814, #1831, #1845, #2006), each with a matching infrastructure PR
+(#7351, #7468, #7507, #7534). The three docs files carry the same fact
+three times, and every other system already knows the answer: the release
+registry knows which minors exist, Terraform knows which hosts exist, and
+the legacy repo knows which snapshots exist.
+
+The cut commit is not really a judgment call either: `v7_7` was branched
```
enriquegh, line 30: In this case, I think we picked the wrong commit. If we see 7.7.sourcegraph.com we see ABC now as GA where we should have probably picked f0addc6 where we still callout ABC as beta

dev/versioning-proposal.md
```diff
@@ -0,0 +1,134 @@
...
+triggers step 2.
+
+### 4. DNS: one infrastructure PR per major, not per minor
+
+Add the whole next major to `local.major_versions` at once (`8.0` … `8.12`).
+An unclaimed CNAME to `cname.vercel-dns.com` serves Vercel's 404 until step
+2 maps it. The `versioned-docs-robots` Cloudflare worker reads the same
+list, so the robots rule is covered too. This turns an Entitle grant plus PR
+every month into one per year; a bot with standing write access to
+`infrastructure` is the alternative and probably not worth the access.
```
enriquegh, line 110: I guess its not the end of the world if there's a unclaimed CNAME? Maybe Sec team has an opinion

dev/versioning-proposal.md
```diff
@@ -0,0 +1,134 @@
...
+  needs no metadata commit at all: cut from the parent of the bump, its list
+  already ends at `P.Q`.
+
+Why the list stays in the repo instead of being fetched from the release
+registry at build time: `/v1/releases/sourcegraph` returns the newest 50
+releases and ignores `limit`, `page` and `offset`, so it reaches 6.4 today
+and loses a minor every few patch releases. A frozen host also has to pin
+its own version, which the registry cannot tell it.
+
+### 2. Archive the previous version on merge (docs repo, two secrets)
```
enriquegh, line 71: See above but I think this is the biggest part of this. We don't really have a "develop" or "next" branch we consistently use for content that is coming up for the next version.

This means if we merge some cool new feature for 8.1 before the release actually happened, we'll now have that in the "old" 8.0.sourcegraph.com.

dev/versioning-proposal.md
enriquegh: fixed that ^ btw

